PR #39, opened 2018-10-31
Applying cas value present in Document struct

rcashie:
...to allow for atomic storage operations.


daschl:
Thanks for your contribution! We have changed the codebase quite a bit, so this PR unfortunately does not apply anymore. But here is the new equivalent that considers your change 
5a654aae233c6e44e0d9f666ef869c616415d252

thanks much again!

Issue #45, opened 2019-07-08
Static Build with OpenSSL


tlebovic:
I am trying to statically build my executable with musl. I need to be able to provide couchbase with a new location where I have compiled OpenSSL. I noticed that there is a `with-openssl` flag that can be used for that very purpose. However, I don't believe it's exposed when compiling couchbase-rs as a project dependency. Probably missing something very simple so apologies ahead of time.

daschl:
You are right I don’t think this works yet - I haven’t had that come up before.

I‘ll check with @avsej and report back

daschl:
He said libcouchbase accepts the path as `cmake -DOPENSSL_ROOT_DIR=/path` so I think we make the build.rs take an env variable and pass it down this might be an option @tlebovic?

tlebovic:
Yeah, making build.rs expose that option through an environment variable would work nicely. If you want I can submit a PR.

daschl:
@tlebovic thanks for the PR! Let's continue the discussion over there on the specific implementation.

daschl:
A similar build.rs change is merged, so I think this can be closed. Thanks for all your help on this!

PR #46, opened 2019-07-11
Exposed libcouchbase environment variables through build.rs mechanism.

tlebovic:
I have created this PR based on the discussion in #45.

### DIFF
--- a/couchbase-sys/build.rs
+++ b/couchbase-sys/build.rs
@@ -3,6 +3,18 @@
 use std::env;
 use std::path::PathBuf;
 
+static ENV_FLAG_PREFIX: &'static str = "CB_";
+
+// Simple Environment Vairable filter
+fn environment_variable_filter(env_flag:&str) -> bool {
+    if env_flag.len() > ENV_FLAG_PREFIX.len() &&
+       &env_flag[0..ENV_FLAG_PREFIX.len()] == ENV_FLAG_PREFIX {
+        true
+    } else {
+        false
+    }
+}
+
 fn main() {
     let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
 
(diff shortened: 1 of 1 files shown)


daschl:
@tlebovic this looks great, thanks! One idea that we can talk through briefly:

What about if instead of enumerating all lcb variables we iterate through all env variables and only pass down that have a common prefix like `LCB_`? The upside would be that we do not need to keep them in sync and users can pass down whatever they need, the downside is since some of them already have `LCB_` in the name you would have to end up passing something like `LCB_LCB_USE_ASAN` - or we find a different prefix? 

In any case I am okay with the change as is, I just wanted to propose a different way which might save us maintenance to keep them in-sync in the future.

tlebovic:
I like the idea, just not a big fan of the possible double "LCB_" on some environment variables. Maybe the prefix should be "CBRS_"? The only other downside I can see with this approach is if you ever want to restrict users from setting specific build settings.

daschl:
@tlebovic we could just use "CB_" ? .. I think if we ever want to restrict users we can blacklist them?

using "CB_" we can also open the door to add our own rust flags in the future, so it blends with the lcb ones.

tlebovic:
@daschl that sounds good to me. I will submit changes probably tomorrow evening based on this discussion. Perhaps we will not implement the blacklisting right now, but it shouldn't be hard to add in the future if we need it.

couchbase-sys/build.rs
```diff
@@ -3,6 +3,18 @@
 use std::env;
 use std::path::PathBuf;
 
+static ENV_FLAG_PREFIX: &'static str = "CB_";
+
+// Simple Environment Vairable filter
+fn environment_variable_filter(env_flag:&str) -> bool {
+    if env_flag.len() > ENV_FLAG_PREFIX.len() &&
```
daschl: can this be simplified into just `env_flag.len() > ENV_FLAG_PREFIX.len() &&
       &env_flag[0..ENV_FLAG_PREFIX.len()] == ENV_FLAG_PREFIX` ? (avoid the if and return true false which is I think implicit in the logical and)

daschl:
@tlebovic btw, have you signed the couchbase CLA? if not, can you please take a minute and follow the initial steps under https://github.com/couchbase/couchbase-java-client/blob/master/CONTRIBUTING.md#contributing ? Should not take long and once this is done you do not need to do it again.

Thanks much again for your great input and contributions!

couchbase-sys/build.rs
tlebovic: Yeah, I can make the change. The only reason I had it in the if/else block was to allow for whitelisting and subsequent filtering. But can remove it for now.

daschl:
@tlebovic heyo let me know if you are stuck on things here to help you out

tlebovic:
Sorry, got pulled off and didn't get the chance to circle back sooner. I will also complete the CLA in the next couple of days.

sadolit:
Hi, are there any plans to continue with the PR?

daschl:
Yes, I was waiting if @tlebovic was
Going to sign the CLA - once that is done we can merge it!

daschl:
Since this is a rather contained change, @tlebovic for the interest of progress, if I don't get a response I'll just close this one out and get a similar change in under my name. I'll mention you in the commit if that's alright.

daschl:
4da22840617eb8806a96c35e5576026fae170280 , thanks @tlebovic !

Issue #21, opened 2017-02-22
Remove all unwrap and expect calls where possible


daschl:
Lets make sure we build a well behaving and safe client at runtime.

adamhjk:
PR #48 removes all the expect calls other than at the C interop layer.

daschl:
@adamhjk thanks much!

PR #48, opened 2019-10-15
feature: Convert to async/await

adamhjk:
This commit switches from futures 0.1 to the new async/await syntax. It
requires Rust 1.39+, currently in beta. Additionally, it switches from
`std::sync::Mutex` to the futures-safe Mutex from `futures-preview`.
This should solve deadlock problems when running on multi-threaded
futures executors, such as tokio.

All tests pass, and the rustdoc is updated.

### DIFF
--- a/couchbase/src/bucket.rs
+++ b/couchbase/src/bucket.rs
@@ -7,6 +7,7 @@ use std::rc::Rc;
 use std::sync::Arc;
 
 /// Provides access to `Bucket` level operations and `Collections`.
+#[derive(Debug)]
 pub struct Bucket {
     instance: Rc<Instance>,
 }
@@ -64,6 +65,7 @@ impl Bucket {
 }
 
 /// Provides access to `SharedBucket` level operations and `SharedCollections`.
+#[derive(Debug)]
 pub struct SharedBucket {
     instance: Arc<SharedInstance>,
 }
@@ -97,7 +99,7 @@ impl SharedBucket {
     where
         S: Into<String>,
     {
-        self.instance.query(statement.into(), options).await.await
+        self.instance.query(statement.into(), options).await
     }
 
     /// Internal proxy method that gets called from the cluster so we can send it into the
@@ -110,7 +112,7 @@ impl SharedBucket {
     where
         S: Into<String>,
     {
-        self.instance.analytics_query(statement.into(), options).await.await
+        self.instance.analytics_query(statement.into(), options).await
     }
 
     /// Internal proxy method that gets called from the cluster so we can send it into the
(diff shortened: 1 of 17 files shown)


adamhjk:
This also now removes all the panics other than in the C interop layer, resolving #21

adamhjk:
Done - my CLA is on file in Gerrit.

adamhjk:
Also - sorry it's a bit of a code bomb. I couldn't see a more incremental way to get it done. :)

couchbase/src/bucket.rs
```diff
@@ -33,28 +33,28 @@ impl Bucket {
...
         S: Into<String>,
     {
-        self.instance.query(statement.into(), options)
+        self.instance.query(statement.into(), options).await
     }
 
     /// Internal proxy method that gets called from the cluster so we can send it into the
     /// instance.
-    pub(crate) fn analytics_query<S>(
+    pub(crate) async fn analytics_query<S>(
```
daschl: There are many methods like this so I'll just take this one to discuss the pattern:

I'm unsure if these "forward" methods  gain anything from being converted to async await, since we are not actually awaiting on something? I'm somewhat cautious that the async/await sugar for these method introduces some (unnecessary?) overhead that is not needed since it is just forwarding the underlying future result?

On the other hand I have no data to back that feeling up and I am also not sure what the best practice is going to be on that. What's your take on this?

daschl: Other than that, looks pretty great to me :)

adamhjk: I struggled with this! I'm not an expert on async/await - but here is why I landed where I did. Essentially, when you declare a function as `async`, what Rust does is generate the future state machine for you, based on all the calls to `await` within the function. 

Compared to futures 0.1, there is a little more overhead (you can see this in some of the outstanding issues around async/await - in particular, things can get a little out of control if you pass futures as arguments to functions, or worse, if you do it recursively). I'm pretty sure that it won't be a big differential, either in performance or resource utilization. 

As a bit of practicality, when I "passed through" the future generated from the real function, rather than the forward one, I wound up with an internal API that had a lot of calls to `await.await`- which felt like a code smell. In an async/await world, I think the right thing to do is let the compiler do the work.

adamhjk: In practice, these functions are async - they return the future from upstream when it is available. Nothing actually happens, of course, until they are passed to an executor.

adamhjk: I checked in with the tokio discord - there shouldn't be an advantage to returning the raw future, unless the intent is to build the state machine by hand (which we certainly don't want).

daschl: @adamhjk thanks much for checking, looks good to me then!

daschl:
Thanks so much @adamhjk for this contribution!

adamhjk:
My pleasure! Thanks for the quick response.

PR #55, opened 2019-12-10
Scan consistency

### LINKED ISSUE #53: n1ql scan consistency
It appears that we can't set the n1ql `scan_consistency` parameter, meaning that every request can just be stale. This is sometimes fine (and fast!) but is problematic when you need to read-your-own-writes.

https://docs.couchbase.com/server/6.5/settings/query-settings.html#scan_consistency

adamhjk:
This fixes #53, along with exposing the query meta struct fields as public.

It does not implement all the possible query options; seems like we should let the C API stabilize first, and then do it all at once.

### DIFF
--- a/couchbase/src/options.rs
+++ b/couchbase/src/options.rs
@@ -172,22 +172,45 @@ impl RemoveOptions {
     }
 }
 
+// This maps to LCB_N1QL_CONSISTENCY enum in couchbase.h
+#[repr(u32)]
+#[derive(Debug, Clone, Copy)]
+pub enum ScanConsistency {
+    // LCB_N1QL_CONSISTENCY_NONE
+    NotBounded = 0,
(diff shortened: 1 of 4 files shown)


couchbase/src/options.rs
```diff
@@ -172,24 +172,47 @@ impl RemoveOptions {
     }
 }
 
+// This maps to LCB_N1QL_CONSISTENCY enum in couchbase.h
+#[repr(u32)]
+#[derive(Debug, Clone, Copy)]
+pub enum ScanConsistency {
```
daschl: To align with SDK 3, please name it QueryScanConsistency (since there are other scan consistency enums for analytics, views, etc.)

adamhjk: On it!

daschl:
@adamhjk as part of a larger cleanup on master I already got this in -- thanks much for this change :)

PR #61, opened 2020-04-05
Use the string length when decoding error contexts

jensenn:
I came across a panic when trying to test replacing a kv item when another thread updated the key. To do this, I was setting cas on ReplaceOptions. When the failure occured, I saw the following panic:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 12, error_len: Some(1) }'

...

couchbase-rs/couchbase/src/api/mod.rs:254:9
```

This is caused because the code was expecting a null terminated CStr, however, the strings may not have a null terminator. In my case, the key length was 12 characters long and the Utf8Error occurred on character 13, which was not a null character.

The fix (I think) is to use the length when constructing the Rust string.

### DIFF
--- a/couchbase/src/io/lcb/callbacks.rs
+++ b/couchbase/src/io/lcb/callbacks.rs
@@ -11,6 +11,7 @@ use std::ffi::CStr;
 use std::os::raw::{c_char, c_int, c_uint, c_void};
 use std::ptr;
 use std::slice::from_raw_parts;
+use std::str;
 
 use crate::io::lcb::{
     bucket_name_for_instance, decrement_outstanding_requests, wrapped_vsnprintf, AnalyticsCookie,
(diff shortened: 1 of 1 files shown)


daschl:
@jensenn  thanks much for your contribution! Looks good to me, all you need to do is please sign the couchbase CLA (see the first paragraph here https://github.com/couchbase/couchbase-java-client/blob/master/CONTRIBUTING.md#contributing)

daschl:
@jensenn did you sign the CLA? I can also get it in but I'd like to give you credit in the github commit history

daschl:
To make progress on this I'll close it and get it merged in a different way -  thanks for letting us know that this is an issue!

PR #63, opened 2020-04-23
Add accessors for cas/mutation token to MutationResult

chvck:


couchbase/src/api/results.rs
```diff
@@ -258,6 +258,14 @@ impl MutationResult {
             mutation_token,
         }
     }
+
+    pub fn cas(&self) -> u64 {
+        self.cas
+    }
+
+    pub fn mutation_token(&self) -> &Option<MutationToken> {
```
brett19, line 266: Based on my admittedly lacking Rust knowledge, I believe this signature should actually be:
`pub fn mutation_token(&self) -> Option<&MutationToken>`

daschl:
Looks good!

PR #67, opened 2020-06-03
Add minimal implementation of search query

chvck:


couchbase/examples/search.rs
```diff
@@ -0,0 +1,34 @@
...
+    env_logger::init();
+
+    // Connect to the cluster with a connection string and credentials
+    let cluster = Cluster::connect("couchbase://127.0.0.1", "Administrator", "password");
+    // Open a bucket
+    let bucket = cluster.bucket("travel-sample");
+    // Use the default collection (needs to be used for all server 6.5 and earlier)
+    let _collection = bucket.default_collection();
+
+    // Perform a simple analytics query
```
daschl, line 19: copy/paste typo

couchbase/examples/search.rs
```diff
@@ -0,0 +1,34 @@
...
+    match block_on(cluster.search_query::<QueryStringQuery>(
```
daschl, line 20: Do you need to provide the generic type here even if you pass it in the arguments?

couchbase/src/api/mod.rs
```diff
@@ -155,6 +156,58 @@ impl Cluster {
         receiver.await.unwrap()
     }
 
+    /// Executes an analytics query
```
daschl, line 159: Most of this doc block still has analytics stuff in it

couchbase/src/api/mod.rs
```diff
@@ -155,6 +156,58 @@ impl Cluster {
...
+    ///             println!("Found Row {:?}", row);
+    ///         }
+    ///     },
+    ///     Err(e) => panic!("Query failed: {:?}", e),
+    /// }
+    /// ```
+    /// See the [SearchResult](struct.SearchResult.html) for more information on what and how it can be consumed.
+    pub async fn search_query<T: SearchQuery>(
+        &self,
+        index: String,
```
daschl, line 197: similar to the other options, for convenience, make this take a generic S where S: Into<String> .. then inside the code do index.into() -- if someone passes in a &str it will be cloned into an owned one, if it is a String already it is a noop

couchbase/src/api/results.rs
```diff
@@ -169,6 +170,49 @@ pub struct AnalyticsMetaData {
     client_context_id: String,
 }
 
+#[derive(Debug, Deserialize)]
+pub struct SearchMetaData {
+    errors: Option<HashMap<String, String>>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SearchRow {
```
daschl, line 179: Is this used at all in your code, if you turn it right into a json value ?

chvck: As discussed, used in the result.

PR #71, opened 2020-06-17
Add support for bucket ping

chvck:


couchbase/src/api/results.rs
```diff
@@ -457,3 +457,121 @@ impl KvStat {
...
+
+#[derive(Debug, Hash, Eq, PartialEq, Clone, Copy)]
+pub enum PingState {
+    OK,
+    Timeout,
+    Error,
+    Invalid,
+}
+
+impl fmt::Display for PingState {
```
daschl, line 573: that's a bit odd - any reason why you are basically implementing Debug for display?

couchbase/src/io/lcb/encode.rs
```diff
@@ -697,3 +698,23 @@ pub fn encode_kv_stats(instance: *mut lcb_INSTANCE, request: KvStatsRequest) {
         lcb_stats3(instance, cookie as *mut c_void, &command);
     }
 }
+
+/// Encodes a `PingRequest` into its libcouchbase `lcb_CMDPING` representation.
+pub fn encode_ping(instance: *mut lcb_INSTANCE, request: PingRequest) {
+    let cookie = Box::into_raw(Box::new(request.sender));
+
+    let report_id = match request.options.report_id {
```
daschl, line 706: This can be simplified to something like unwrap_or() on the optional.. if you are worried about the uuid allocation, there is also a variant IIRC which takes a callback

couchbase/src/api/results.rs
chvck: I just wanted to expose a way to call to_string() on the enum so that it can be printed out

PR #73, opened 2020-07-16
Add support for building on Windows

chvck:
We should probably test that linux/macos builds still work...


couchbase-sys/build.rs
```diff
@@ -50,17 +50,35 @@ fn main() {
         build_cfg.define("LIBCOUCHBASE_STATIC", "ON");
     }
 
+    if cfg!(any(target_os = "windows")) {
```
daschl, line 53: I don't think you need the any with a single entry.. I think cfg!(target_os = "windows") should work?

couchbase-sys/build.rs
```diff
@@ -50,17 +50,35 @@ fn main() {
...
+
     let build_dst = build_cfg.build();
 
     if cfg!(feature = "link-static") {
-        std::fs::copy(
-            format!("{}/libcouchbaseS.a", build_dst.join("build/lib").display()),
-            format!("{}/libcouchbase.a", build_dst.join("build/lib").display()),
-        )
-        .unwrap();
+        if cfg!(any(target_os = "windows")) {
```
daschl, line 60: same as above

couchbase-sys/build.rs
```diff
@@ -50,17 +50,35 @@ fn main() {
...
+            std::fs::copy(
+                format!("{}/libcouchbaseS.a", build_dst.join("build/lib").display()),
+                format!("{}/libcouchbase.a", build_dst.join("build/lib").display()),
+            )
+            .unwrap();
+        }
 
         if cfg!(any(target_os = "macos", target_os = "freebsd")) {
             println!("cargo:rustc-link-lib=dylib=c++");
+        } else if cfg!(any(target_os = "windows")) {
```
daschl, line 79: same as above ;)

PR #93, opened 2020-11-19
copy precompiler check from cbsasl/src/client.c

avsej:


chvck:
This will be added to lcb but should give us a headstart on diagnosing an issue in cbsh

PR #99, opened 2021-01-26
Add support for collections manager

chvck:


couchbase/src/api/collections.rs
```diff
@@ -0,0 +1,324 @@
...
+impl ScopeSpec {
+    pub fn new(name: String, collections: Vec<CollectionSpec>) -> Self {
+        Self { name, collections }
+    }
+
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn collections(&self) -> &Vec<CollectionSpec> {
```
daschl, line 30: you could return a slice here?

couchbase/src/api/collections.rs
```diff
@@ -0,0 +1,324 @@
...
+use std::time::Duration;
+
+#[derive(Debug)]
+pub struct ScopeSpec {
+    name: String,
+    collections: Vec<CollectionSpec>,
+}
+
+impl ScopeSpec {
+    pub fn new(name: String, collections: Vec<CollectionSpec>) -> Self {
```
daschl, line 22: if these are public, for convenience it could make sense to not accept string but rather S: Into<String>, then someone can either provide a string or a &'static str like ::new("foo")

couchbase/src/api/collections.rs
```diff
@@ -0,0 +1,324 @@
...
+            }
+            scopes.push(ScopeSpec::new(scope.name, collections));
+        }
+
+        Ok(scopes)
+    }
+
+    pub async fn create_scope(
+        &self,
+        scope_name: String,
```
daschl, line 147: same into string strategy on all public methods?

daschl (approved):
+2, couple nits maybe you want to address them, not a biggie if we change it later

daschl (approved):

PR #102, opened 2021-01-28
Remove ram from bucket builder new

chvck:


couchbase/src/api/buckets.rs
```diff
@@ -177,10 +177,10 @@ pub struct BucketSettingsBuilder {
 }
 
 impl BucketSettingsBuilder {
-    pub fn new<S: Into<String>>(name: S, ram_quota_mb: u64) -> BucketSettingsBuilder {
+    pub fn new<S: Into<String>>(name: S) -> BucketSettingsBuilder {
         Self {
             name: name.into(),
-            ram_quota_mb,
+            ram_quota_mb: 0,
```
daschl, line 183: should we set the default to 100 for simplicity?

chvck: Yeah I like that idea

daschl (approved):

PR #140, opened 2022-04-17
Tried to use couchbase in rust, found out that your cmake reference didn't support VISUAL STUDIO 2022

TiloPapke:
The depedency of cmake ist updated to 0.1.48 so that Visual Studio 2022 is support.

still unable to compile your code example "couchbase.lib" could nocht be opened


chvck:
Closing as the version of the SDK using lcb is no longer in development.

PR #142, opened 2022-12-07
fix: `__va_list_tag` not found

kkharji:
## Overview 

This PR solves to issues I faced during building and compiling `couchbase-rs` on macos M1 and rust 1.65.0:

- `__va_list_tag` not found in this scope
- `serde::Serialize` and `serde_derive::Serialize` naming conflict

## Details

When trying to build `couchbase-rs` on macbook m1 or arm64-based docker image. The build process errors with: 

```bash
error[E0412]: cannot find type `__va_list_tag` in this scope
    --> couchbase/src/io/lcb/callbacks.rs:1035:31
     |
1035 | pub(crate) type VaList = *mut __va_list_tag;
     |                               ^^^^^^^^^^^^^ not found in this scope
```

Additionally, when trying to integrate `couchbase-rs` in a rust project, I get Serialize macro/struct name conflict errors.

## Solution

1. **Guard `VaList` definition with extra `target_os` and `target_arch` conditions**
    If `target_arch` is either "arch64" or "arm64" then `VaList` will be defined as `va_list`.
    else it should not effect the original configurations.
2. **Expect `u8` or `i8` while building for arm64**
   if `target_arch` == "aarch64" and it's linux then `u8` else `i8`

cc @chvck

### DIFF
--- a/couchbase/src/api/search.rs
+++ b/couchbase/src/api/search.rs
@@ -1,4 +1,3 @@
-use serde::Serialize;
 use serde_derive::Serialize;
 use serde_json::json;
 use std::fmt::Debug;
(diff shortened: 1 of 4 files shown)


chvck:
Thanks for your contribution @kkharji.

PR #148, opened 2024-01-25
Rust sdk upgrade

oltraj:
Upgraded the Rust SDK


couchbase-sys/Cargo.toml
```diff
@@ -21,7 +21,7 @@ cc = "1.0"
 cmake = "0.1.45"
 
 [build-dependencies.bindgen]
-version = "0.55"
+version = "0.62"
```
chvck, line 24: This should probably be an update to the latest version.

couchbase-sys/Makefile
```diff
@@ -0,0 +1,31 @@
+binding:
```
chvck, line 1: For my understanding why do we need this makefile? `cargo build` should run all the bindgen bits automatically.

couchbase-sys/headers-3.3.10.h
```diff
@@ -0,0 +1,24 @@
+#include "libcouchbase-3.3.10/include/libcouchbase/couchbase.h"
```
chvck, line 1: Again for my understanding what does this file do?

couchbase-sys/src/lib.rs
```diff
@@ -8,3 +8,9 @@
 extern crate openssl_sys;
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
```
chvck, line 11: Was this meant to be here?

chvck (changes requested):
I think that the commits for this PR should be tidied up. One for updating bindgen version and then another for updating libcouchbase.

couchbase-sys/Cargo.toml
oltraj: I agree

oltraj:
> I think that the commits for this PR should be tidied up. One for updating bindgen version and then another for updating libcouchbase.

Thank you for your feedback, @chvck. I appreciate your input. I agree that it would be a good idea to tidy up the commits for this PR. I'll create one commit for updating bindgen to the latest version and another for updating libcouchbase. I'll work on this and update the PR accordingly."

couchbase-sys/src/lib.rs
oltraj: Indeed, there appears to be a duplication of the extern crate openssl_sys; and include!(concat!(env!("OUT_DIR"), "/bindings.rs")); lines in the code. This duplication seems unintentional and can be safely removed. The primary purpose of this code is to set up attributes, dependencies, and include generated Rust bindings. Thank you for pointing this out, and we will make sure to clean it up in the code.

chvck: Sorry, I was referring to the commented out unit test below.

oltraj:
> I think that the commits for this PR should be tidied up. One for updating bindgen version and then another for updating libcouchbase.

Created the PR for the two : 
* https://github.com/couchbaselabs/couchbase-rs/pull/150
* https://github.com/couchbaselabs/couchbase-rs/pull/151

oltraj:
Hello @chvck  , please review the created two PR, thank you

chvck:
Closing as this work has now been completed, and merged, in another PR.

PR #152, opened 2024-04-15
Update to Rust v1.77.2 and libcouchbase v3.3.12

ayushmishra2005:


ayushmishra2005:
> Hi @ayushmishra2005 thanks, this looks good to me. Before we merge I think that we should fix CI though so that we can verify the change works on all platforms. I made a start at #153 but it needs some investigation what's going on with the Windows build.

Hi @chvck Thanks for raising #153 . I don't have a window machine but I will try to take a look and let you know if I find anything.

chvck:
Looks like the problem might just be some Windows specific code.

ayushmishra2005:
> Looks like the problem might just be some Windows specific code.

Yes @chvck you are right. I just pushed some changes to fix this. Could you please approve workflow?

chvck:
This is looking good @ayushmishra2005 could you just update the commits to combine the similar commits; 1 for updating the readme, 1 for fixing tests, and 1 for refactoring please?

PR #155, opened 2024-04-30
Added test-case infrastructure and tests for collection and bucket

ayushmishra2005:


couchbase/src/tests/collection.rs
```diff
@@ -0,0 +1,723 @@
...
+    );
+    let result: CouchbaseResult<()> = mocked_collection
+        .unlock(key, 1, UnlockOptions::default())
+        .await;
+
+    assert!(result.is_ok());
+}
+
+#[tokio::test]
+async fn lookup_in_works() {
```
chvck, line 607: Is there much value to these tests that just check that the Cas value is 0? Relatedly, the Cas value should never be 0 on a successful response from the server.

couchbase/src/io/request.rs
```diff
@@ -29,6 +29,11 @@ pub enum Request {
     Touch(TouchRequest),
     GetReplica(GetReplicaRequest),
 }
+impl PartialEq for Request {
+    fn eq(&self, other: &Request) -> bool {
+        true
```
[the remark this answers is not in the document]
ayushmishra2005: @chvck I had to do this to pass request in `mockall` expectation https://github.com/ayushmishra2005/couchbase-rs/blob/0b61447730eb52e6a7f9b8dbe327fd44e85ea23d/couchbase/src/tests/collection.rs#L43

I have just made a correction and made it only test-specific.

couchbase/src/tests/collection.rs
ayushmishra2005: I have just corrected it. Please review

couchbase/src/tests/collection.rs
```diff
@@ -0,0 +1,723 @@
+use crate::api::keyvalue_options::{GetReplicaOptions, ReplicaMode};
```
chvck, line 1: Is there much value to most of these tests? These tests don't actually seem to test any of the forks or the logic in the functions being called, e.g. the get with expiry test doesn't test that the expiry is parsed correctly or that the value is returned either. Would a number of these tests also be better somehow mocking out the LcbInstance in the io module? Or just being integration tests maybe?

In several cases the value being returned by the mockcore also isn't valid for the request. e.g. In the LookupIn test the request asks for a single xattr but gets back 2 non-xattr values completely unrelated to the request.

ayushmishra2005: Hi @chvck 

We already have integration tests to verify integration between different components of the application.

This PR contains only unit tests of collection.rs and bucket.rs. The purpose of these tests is to

1) Test all methods by passing in input and asserting the actual output against an expected one.
2) I have used mocking framework to mock out external  calls but still asserting that the calls are made.

Regarding verification of expected output, let me go through all tests again and modify wherever needed. I will let you know once its done. I just modified expiry test .

chvck: Sounds good @ayushmishra2005 is it also worth testing that the input at the point it reaches the core is also as expected?

ayushmishra2005: Yes @chvck . I am already doing this. [This](https://github.com/ayushmishra2005/couchbase-rs/blob/0b61447730eb52e6a7f9b8dbe327fd44e85ea23d/couchbase/src/tests/collection.rs#L44)  checks that `core.send()` should be called with passed input at least once. If I pass 0 in times, I will panik. Let me add  panic tests for the same.

chvck: Isn't that check just checking that *any* request is sent, rather than the details of the request? Like the request is the correct type and has the correct field values?

ayushmishra2005: Hi @chvck If we pass wrong request as parameter, it will panic. I have added few panic tests for collection. Could you please take a look at this commit https://github.com/couchbaselabs/couchbase-rs/pull/155/commits/b28d11de5067f4ea155a38e7a9f7124a732feed2 


I have added here https://github.com/ayushmishra2005/couchbase-rs/blob/b28d11de5067f4ea155a38e7a9f7124a732feed2/couchbase/src/tests/collection.rs#L164


Please let me know if it makes sense. If yes, I will add panic tests for other functions.

PR #158, opened 2024-07-10
Couchbase2-core/Protostellar initial commit

Matt-Woz:
- Added Protostellar package w/ proto autogen
- initial commit w/ couchbase2-core package


.github/workflows/build.yml
```diff
@@ -42,6 +44,18 @@ jobs:
         run: sudo apt-get update && sudo apt-get install gcc-aarch64-linux-gnu -y
         if: matrix.target == 'aarch64-unknown-linux-gnu'
 
+      - name: Install protoc
```
chvck, line 47: If we're checking in the generated files then do we need protoc?

Matt-Woz: No, I don't think so. I removed it from GH actions and moved the proto generation out of the build script into a separate binary

chvck (approved):

PR #168, opened 2024-07-30
Update memdx channel usage to use callbacks where possible

chvck:
Motivation
-----------
Using channels for things like orphan handler requires us to spin up new tasks, the flow is also a bit easier to follow with callbacks.

Changes
-------
Update orphan handling and connection close handling to use callbacks. Rewrite kvclientpool to make this possible.


Matt-Woz (approved):
LGTM!

PR #177, opened 2024-08-21
RSCBC-1: Implement HTTP Crate

Matt-Woz:
Changes
---------
- Created HTTP Client, wrapping reqwest
- Created custom JSON Row parser + stream

### DIFF
--- a/sdk/couchbase-core/src/httpx/base.rs
+++ b/sdk/couchbase-core/src/httpx/base.rs
@@ -1,0 +1,125 @@
+use async_trait::async_trait;
+use bytes::Bytes;
+use futures::{Stream, StreamExt};
+use http::{StatusCode};
+use serde::de::DeserializeOwned;
+use typed_builder::TypedBuilder;
+use crate::httpx::error::ErrorKind::Generic;
+use crate::httpx::error::{Result as HttpxResult};
+
+#[derive(Debug, TypedBuilder)]
+#[builder(field_defaults(default, setter(into)))]
+#[non_exhaustive]
+pub struct Request {
+    pub method: http::Method,
+    pub uri: String,
+    pub auth: Option<Auth>,
+    pub user_agent: Option<String>,
+    pub content_type: Option<String>,
+    pub body: Option<String>,
+}
+
+#[derive(PartialEq, Eq, Debug, Clone, TypedBuilder)]
+#[builder(field_defaults(default, setter(into)))]
+pub struct BasicAuth {
+    pub username: String,
+    pub password: String,
+}
+
+#[derive(PartialEq, Eq, Debug)]
+#[non_exhaustive]
+pub enum Auth {
+    BasicAuth(BasicAuth),
+    OnBehalfOf(OnBehalfOfInfo)
+}
+
+#[derive(PartialEq, Eq, Debug, TypedBuilder)]
+#[builder(field_defaults(setter(into)))]
+pub struct OnBehalfOfInfo {
+    pub username: String,
+    pub password_or_domain: OboPasswordOrDomain,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum OboPasswordOrDomain {
+    Password(String),
+    Domain(String),
+}
+
+#[derive(PartialEq, Eq, Debug, TypedBuilder)]
+#[builder(field_defaults(default, setter(into)))]
+#[non_exhaustive]
+pub struct RequestCreator {
+    endpoint: String,
+    user_agent: Option<String>,
+    basic_auth: Option<BasicAuth>,
+}
+
+#[derive(Debug)]
+pub struct Response<R> {
+    pub inner: R,
+}
+
+#[async_trait]
+pub trait ResponseProvider {
+    fn get_status(self) -> StatusCode;
+
+    async fn get_text(self) -> HttpxResult<String>;
+
+    fn get_stream(self) -> Box<dyn Stream<Item = HttpxResult<Bytes>> + Unpin>;
+
+    async fn read_response_as_json<T: DeserializeOwned>(self) -> HttpxResult<T>;
+}
+
+#[async_trait]
+impl ResponseProvider for reqwest::Response {
+    fn get_status(self) -> StatusCode {
+        self.status()
+    }
+    async fn get_text(self) -> HttpxResult<String> {
+        let text = self.text().await?;
+        Ok(text)
+    }
+
+    fn get_stream(self) -> Box<dyn Stream<Item=HttpxResult<Bytes>> + Unpin> {
+        let x = self.bytes_stream().map(|result| {
+            result.map_err(|e| Generic { msg: e.to_string() }.into())
+        });
+        Box::new(x)
+    }
+
+    async fn read_response_as_json<T: DeserializeOwned>(self) -> HttpxResult<T> {
+        self.json::<T>()
+            .await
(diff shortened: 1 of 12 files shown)


sdk/couchbase-core/src/httpx/json_row_stream.rs
```diff
@@ -0,0 +1,79 @@
...
+where
+    S: Stream<Item = Result<Bytes, reqwest::Error>> + Unpin,
+{
+    type Item = HttpxResult<Vec<u8>>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<HttpxResult<Vec<u8>>>> {
+        let this = self.get_mut();
+        loop {
+            match this.state {
+                State::Collecting => match this.parser.next() {
```
chvck, line 56: Nice! I think a few comments explaining each case here would be v. helpful.

sdk/couchbase-core/src/httpx/raw_json_row_streamer.rs
```diff
@@ -0,0 +1,360 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::httpx::base::{Body, Method, RequestBuilder};
+    use crate::httpx::client::Client;
+    use crate::httpx::json_row_stream::JsonRowStream;
+    use crate::httpx::raw_json_row_streamer::RawJsonRowStreamer;
+    use serde_json::{json, Value};
+
+    #[tokio::test]
+    async fn stream_some_query_rows() {
```
chvck, line 214: Same comment as the other test.

chvck:
This is great. It might be a good idea to move the integration tests to the top level `tests` crate and think about how we can add an abstraction layer between us and reqwest. That'd both decouple us a bit and let us unit test more easily.

sdk/couchbase-core/src/httpx/base.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) struct Body {
+    pub inner: reqwest::Body,
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub(crate) struct OnBehalfOfInfo {
+    username: String,
```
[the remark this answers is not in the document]
Matt-Woz: Yep - done

sdk/couchbase-core/src/httpx/base.rs
```diff
@@ -0,0 +1,123 @@
...
+    password: String,
+    domain: String,
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub(crate) struct RequestBuilder {
```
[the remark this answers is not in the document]
Matt-Woz: Didn't know we had that! Done

sdk/couchbase-core/src/httpx/base.rs
```diff
@@ -0,0 +1,123 @@
...
+            user_agent,
+            endpoint,
+            basic_auth_user,
+            basic_auth_pass,
+        }
+    }
+
+    pub(crate) fn new_request(
+        &self,
+        client: Client,
```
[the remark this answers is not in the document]
Matt-Woz: Yes it is a bit weird, reqwest recommends to use the RequestBuilder, which can be only accessed via a Client, and then running `RequestBuilder.send`, but I refactored to make it more intuitive

sdk/couchbase-core/src/httpx/client.rs
```diff
@@ -0,0 +1,40 @@
...
+        match req.inner.send().await {
+            Ok(response) => Ok(Response { inner: response }),
+            Err(err) => {
+                if err.is_connect() {
+                    Err(Connect {
+                        msg: err.to_string(),
+                    }
+                    .into())
+                } else {
+                    Err(Generic {
```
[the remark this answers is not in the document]
Matt-Woz: Ack

sdk/couchbase-core/src/httpx/json_block_read.rs
```diff
@@ -0,0 +1,79 @@
...
+        compat_version: String,
+        #[serde(alias = "clusterUUID")]
+        cluster_uuid: String,
+        #[serde(alias = "isBalanced")]
+        is_balanced: bool,
+        orchestrator: String,
+    }
+
+    #[tokio::test]
+    async fn get_cluster_info() {
```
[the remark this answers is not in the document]
Matt-Woz: I've moved the integration tests into tests/

sdk/couchbase-core/src/httpx/json_row_stream.rs
Matt-Woz: Done!

sdk/couchbase-core/src/httpx/base.rs
```diff
@@ -0,0 +1,123 @@
...
+    Put,
+    Delete,
+    Connect,
+    Options,
+    Trace,
+    Patch,
+}
+#[derive(Debug)]
+pub(crate) struct Request {
+    pub inner: reqwest::RequestBuilder,
```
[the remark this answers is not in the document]
Matt-Woz: Yes agreed, I refactored to make it a lot more generic

sdk/couchbase-core/src/httpx/base.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub uri: String,
+    pub basic_auth_user: String,
+    pub basic_auth_pass: String,
+    pub method: http::Method,
+    pub content_type: String,
+    pub on_behalf_of: Option<OnBehalfOfInfo>,
+    pub body: Option<String>,
+}
+
+#[derive(PartialEq, Eq, Debug, TypedBuilder)]
```
chvck, line 22: All of the request type structs probably need to have #[non_exhaustive] on them.

sdk/couchbase-core/src/httpx/client.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Client<C> {
+    pub inner: C,
+}
+
+#[async_trait]
+pub trait ClientProvider {
+    type R: ResponseProvider;
+    fn new(/*tlsConfig*/) -> HttpxResult<Self> where Self: Sized;
```
chvck, line 18: Could we file a rscbc for adding tls to http please?

chvck (commented):
This is looking good to me, just the minor thing about #[non_exhaustive] so that expanding request blocks aren't breaking changes in the future. In the future we might want to think about using hyper but this shields any upstream crates from that too.

sdk/couchbase-core/src/httpx/base.rs
```diff
@@ -0,0 +1,103 @@
...
+use http::{StatusCode};
+use serde::de::DeserializeOwned;
+use typed_builder::TypedBuilder;
+use crate::httpx::error::ErrorKind::Generic;
+use crate::httpx::error::{Result as HttpxResult};
+
+#[derive(Debug, TypedBuilder)]
+#[builder(field_defaults(default, setter(into)))]
+#[non_exhaustive]
+pub struct Request {
```
chvck, line 13: Sorry, I missed this before. Most of these fields in these various structs should probably be Option<> if they are actually optional.

sdk/couchbase-core/src/httpx/client.rs
Matt-Woz: Created RSCBC-44 👍

chvck (approved):

PR #178, opened 2024-08-22
Move crud options to typedbuilder

chvck:


Matt-Woz (approved):
LGTM 👍

sdk/couchbase-core/src/memdx/datatype.rs
```diff
@@ -6,6 +6,12 @@ pub enum DataTypeFlag {
     Xattrs,
 }
 
+impl Default for DataTypeFlag {
```
Matt-Woz, line 9: NIT: Not a blocker, but we could use `#[derive(Default)]` for this

chvck: I don't think that'd work here would it? How would it know which variant is the default value?

PR #184, opened 2024-09-17
RSCBC-16: Implement basic KV ops in core

Matt-Woz:
Changes:
Implemented:

- delete
- get_and_lock
- get_and_touch
- unlock
- touch
- add
- replace
- append
- prepend
- increment
- decrement


sdk/couchbase-core/src/memdx/response.rs
```diff
@@ -403,6 +403,861 @@ impl TryFromClientResponse for GetResponse {
...
+    pub mutation_token: Option<MutationToken>,
+    pub server_duration: Option<Duration>,
+}
+
+impl TryFromClientResponse for PrependResponse {
+    fn try_from(resp: ClientResponse) -> Result<Self, Error> {
+        let packet = resp.packet();
+        let status = packet.status;
+
+        if status == Status::KeyExists {
```
Matt-Woz, line 1011: Gocbcorex also checks that `request.cas > 0` before returning CasMismatch, but we don't have too easy a way to do that here.. is it necessary?

sdk/couchbase-core/src/memdx/error.rs
```diff
@@ -206,8 +206,12 @@ pub enum ServerErrorKind {
     KeyNotFound,
     #[error("Temporary failure")]
     TmpFail,
+    #[error("CAS mismatch")]
+    CasMismatch,
```
Matt-Woz, line 210: Should `CasMismatch` be a `ServerErrorKind`, or something different?

chvck: Seems like a `ServerErrorKind` to me. The only place it can originate is from the server.

sdk/couchbase-core/src/memdx/response.rs
chvck: I'm not 100% sure why gocbcorex has that check. This is another reason why I want to refactor all of this to be callback based so that we have scope on the request at this point. For now I'd say go with this and file an issue to add the cas check later.

chvck (approved):

sdk/couchbase-core/src/memdx/response.rs
Matt-Woz: 👍 Created RSCBC-46

PR #185, opened 2024-09-24
RSCBC-2: Add support for query

chvck:


sdk/couchbase-core/src/queryx/query_options.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub statement: Option<String>,
+    pub timeout: Option<Duration>,
+    pub tx_data: Vec<u8>,
+    pub tx_id: Option<String>,
+    pub tx_implicit: Option<bool>,
+    pub tx_stmt_num: Option<u32>,
+    pub tx_timeout: Option<Duration>,
+    pub use_cbo: Option<bool>,
+    pub use_fts: Option<bool>,
+    pub use_replica: Option<ReplicaLevel>,
```
Matt-Woz, line 135: I suppose this is an enum as opposed to `Option<bool>` for easier serialisation?

Matt-Woz (approved):
LGTM 👍

sdk/couchbase-core/src/queryx/query_options.rs
chvck: I don't think it's bool on the wire and I can envision a world where query adds a new option in the future.

PR #209, opened 2024-11-21
Rework json stream implementation

chvck:
Motivation
----------
We currently implement stream so that it streams all of the rows and then returns None, a user must then cfall read_epilog to finish reading items off the stream - e.g. metadata. This is actually quite challenging to do as a consumer of the json stream, it'd be better if the epilog was parsed as a part of the stream.

### DIFF
--- a/sdk/couchbase-core/src/searchx/error.rs
+++ b/sdk/couchbase-core/src/searchx/error.rs
@@ -1,16 +1,18 @@
 use crate::httpx;
 use http::StatusCode;
-use std::backtrace::Backtrace;
 use std::error::Error as StdError;
 use std::fmt::{Display, Formatter};
+use std::sync::Arc;
+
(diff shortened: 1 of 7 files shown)


sdk/couchbase-core/src/searchx/error.rs
```diff
@@ -1,16 +1,18 @@
...
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
 pub struct Error {
     pub kind: Box<ErrorKind>,
-    pub backtrace: Backtrace,
-    pub source: Option<Box<dyn StdError + 'static + Send + Sync>>,
+
+    // TODO: This shouldn't be arc but I'm losing the will to live.
```
Matt-Woz, line 14: real

Matt-Woz (approved):

PR #210, opened 2024-11-22
RSCBC-27: Implement Subdoc in core

Matt-Woz:


sdk/couchbase-core/src/memdx/client.rs
```diff
@@ -34,15 +34,16 @@ use crate::memdx::error::{CancellationErrorKind, Error, ErrorKind};
...
 use crate::memdx::pendingop::ClientPendingOp;
+use crate::memdx::subdoc::SubdocRequestInfo;
 
 pub(crate) type ResponseSender = Sender<error::Result<ClientResponse>>;
 pub(crate) type OpaqueMap = HashMap<u32, Arc<SenderContext>>;
 
 #[derive(Debug, Clone, Copy)]
 pub struct ResponseContext {
     pub cas: Option<u64>,
+    pub subdoc_info: Option<SubdocRequestInfo>,
```
Matt-Woz, line 45: I don't think there's a way around adding this..

sdk/couchbase-core/src/memdx/error.rs
```diff
@@ -451,10 +478,61 @@ pub enum ServerErrorKind {
     UnknownScopeName,
     #[error("collection name unknown")]
     UnknownCollectionName,
+    #[error("subdoc error")]
```
chvck, line 481: It probably makes sense to make this transparent and just have the message from the underlying error kind?

Matt-Woz: Done

sdk/couchbase-core/src/memdx/response.rs
```diff
@@ -1211,6 +1213,405 @@ impl TryFromClientResponse for DecrementResponse {
...
+            // still considered a success
+        } else if status != Status::Success && status != Status::SubDocMultiPathFailure {
+            return Err(OpsCrud::decode_common_error(
+                resp.packet(),
+                resp.local_addr(),
+                resp.peer_addr(),
+            ));
+        }
+
+        let mut results: Vec<SubDocResult> = Vec::new();
```
chvck, line 1285: Worth using with_capacity rather than new here I think.

sdk/couchbase-core/src/memdx/response.rs
```diff
@@ -1211,6 +1213,405 @@ impl TryFromClientResponse for DecrementResponse {
...
+            while cursor.position() < cursor.get_ref().len() as u64 {
+                if cursor.remaining() < 3 {
+                    return Err(Error::protocol_error("bad value length"));
+                }
+
+                let op_index = cursor
+                    .read_u8()
+                    .map_err(|e| Error::from(ErrorKind::Protocol { msg: e.to_string() }))?;
+
+                if op_index > results.len() as u8 {
```
chvck, line 1537: How can the op_index be greater than results.len()?

sdk/couchbase-core/src/memdx/response.rs
Matt-Woz: Done

sdk/couchbase-core/src/memdx/response.rs
Matt-Woz: Say that we received a value on Operation 0 and 2, but not on Operation 1, during the second iteration: `op_index == 2`, and `results.len() == 1`, so we need to fill all the results in-between with empty SubDocResults

chvck (approved):

PR #211, opened 2024-11-26
RSCBC-62: Refactored hostname handling when parsing configs

chvck:
Motivation
----------
We're currently replacing $HOST in configs with the endpoint reference rather than the source hostname.

Changes
-------
Refactors how we deal with remote hostnames and addresses from the memdx package, to lift addresses in errors up to the core level.
Fix SRV record resolving and connecting to FQDNs.

### DIFF
--- a/sdk/couchbase-connstr/Cargo.toml
+++ b/sdk/couchbase-connstr/Cargo.toml
@@ -6,6 +6,10 @@ edition = "2021"
 [dependencies]
 form_urlencoded = "1.2.1"
 hickory-resolver = { version = "0.24.1", features = ["dns-over-rustls"], default-features = false }
+log = "0.4.22"
 regex = "1.11.0"
 thiserror = "1.0"
 url = "2.5.2"
+
+[dev-dependencies]
+tokio = { version = "1.40", features = ["rt", "net", "macros"] }
--- a/sdk/couchbase-connstr/src/lib.rs
+++ b/sdk/couchbase-connstr/src/lib.rs
@@ -1,6 +1,8 @@
 pub mod error;
 
 use error::ErrorKind;
+use hickory_resolver::TokioAsyncResolver;
+use log::debug;
 use regex::Regex;
 use std::collections::HashMap;
 use std::fmt;
@@ -50,7 +52,7 @@ impl ConnSpec {
             let scheme = scheme_type.as_str();
             if (scheme != "couchbase" && scheme != "couchbases")
                 || self.hosts.len() != 1
-                || self.hosts[0].port.is_none()
+                || self.hosts[0].port.is_some()
             {
                 return None;
             }
@@ -179,7 +181,7 @@ pub struct ResolvedConnSpec {
     pub options: HashMap<String, Vec<String>>,
 }
 
-pub fn resolve(conn_spec: ConnSpec) -> error::Result<ResolvedConnSpec> {
+pub async fn resolve(conn_spec: ConnSpec) -> error::Result<ResolvedConnSpec> {
     let (default_port, has_explicit_scheme, use_ssl) = if let Some(scheme) = &conn_spec.scheme {
         match scheme.as_str() {
             "couchbase" => (DEFAULT_MEMD_PORT, true, false),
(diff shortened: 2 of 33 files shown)


sdk/couchbase-core/src/agent.rs
```diff
@@ -528,13 +530,12 @@ impl Agent {
     fn gen_first_kv_client_configs(
         memd_addrs: &Vec<String>,
         state: &AgentState,
-    ) -> HashMap<String, KvClientConfig> {
+    ) -> Result<HashMap<String, KvClientConfig>> {
```
Peter-Searby, line 533: Why does this need to be a Result? It doesn't look like any error case was added?

sdk/couchbase-connstr/src/lib.rs
```diff
@@ -50,7 +52,7 @@ impl ConnSpec {
             let scheme = scheme_type.as_str();
             if (scheme != "couchbase" && scheme != "couchbases")
                 || self.hosts.len() != 1
-                || self.hosts[0].port.is_none()
+                || self.hosts[0].port.is_some()
```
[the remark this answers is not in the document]
chvck: This just fixes the SRV record resolving. Changing the address handling all the way down into memdx fixes the FQDNs - SocketAddr cannot hold a FQDN - which makes sense really...

sdk/couchbase-core/src/agent.rs
chvck: Good spot, a carry-over from when I was trying to fix the FQDN bits.

sdk/couchbase-connstr/Cargo.toml
```diff
@@ -6,6 +6,10 @@ edition = "2021"
 [dependencies]
 form_urlencoded = "1.2.1"
 hickory-resolver = { version = "0.24.1", features = ["dns-over-rustls"], default-features = false }
+log = "0.4.22"
 regex = "1.11.0"
 thiserror = "1.0"
 url = "2.5.2"
+
+[dev-dependencies]
+tokio = { version = "1.0", features = ["rt", "net", "macros"] }
```
Peter-Searby, line 15: This looks like a very old version of tokio. We seem to have "1.40" and "1.38" in the existing cargo.toml files. I'm not too well informed on the rust dependency management best practices, but this doesn't seem great?

sdk/couchbase-core/src/agent.rs
```diff
@@ -528,13 +530,12 @@ impl Agent {
     fn gen_first_kv_client_configs(
         memd_addrs: &Vec<String>,
         state: &AgentState,
-    ) -> HashMap<String, KvClientConfig> {
+    ) ->HashMap<String, KvClientConfig> {
```
Peter-Searby, line 533: Nit: Looks like some spacing was lost.
I'm honestly quite surprised there's not something like rustfmt being used in this codebase

sdk/couchbase-core/src/configwatcher.rs
```diff
@@ -144,18 +144,21 @@ where
...
-        let addr = client.remote_addr();
-        let host = addr.ip();
+        let hostname = client.remote_hostname();
 
         let resp = client
             .get_cluster_config(GetClusterConfigRequest {})
             .await?;
 
-        let config: TerseConfig = serde_json::from_slice(resp.config.as_slice())?;
+        let config = parse_terse_config(&resp.config, hostname)?;
```
Peter-Searby, line 157: This is a bit confusingly named, considering the call of another function of the same name (albeit a different namespace) immediately after

sdk/couchbase-core/src/agent.rs
chvck: I thought we had it running rustfmt checking on PR, will file an issue to add.

sdk/couchbase-connstr/Cargo.toml
chvck: Not entirely sure why that was 1.0, it's now 1.40.

sdk/couchbase-core/src/configwatcher.rs
chvck: I've added the module name to the call to try to make it more clear. IMO it is the correct name with regards to the module itself though.

Matt-Woz (approved):

PR #212, opened 2024-11-27
Add cargo fmt to github actions

chvck:


.github/workflows/rustfmt.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches:
+      - nativex
+name: Rustfmt
+
+jobs:
+  cargo_fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Rust toolchain and cache
+        uses: actions-rust-lang/setup-rust-toolchain@v1.4.3
```
Peter-Searby, line 13: Out of curiosity, why do we need this for cargo fmt, but the clippy workflow doesn't have it?

Peter-Searby (approved):

Westwooo (approved):

PR #214, opened 2024-12-10
RSCBC-6: Add support for analytics

chvck:


sdk/couchbase/src/options/analytics_options.rs
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, TypedBuilder, Default)]
+#[builder(field_defaults(default, setter(into, strip_option)))]
+#[non_exhaustive]
+pub struct AnalyticsOptions<'a> {
+    pub client_context_id: Option<&'a str>,
+    pub priority: Option<i32>,
```
Matt-Woz, line 28: I think priority should be an `Option<bool>` in the public API?

sdk/couchbase-core/tests/query.rs
```diff
@@ -110,6 +110,7 @@ async fn test_prepared_query_basic() {
     assert_eq!(0, meta.metrics.error_count);
     assert_eq!(0, meta.metrics.warning_count);
 
+    dbg!(&meta.signature);
```
Matt-Woz, line 113: Leftover debug statement

sdk/couchbase/src/options/analytics_options.rs
chvck: Good catch. Done.

sdk/couchbase-core/tests/query.rs
chvck: Done.

Matt-Woz (approved):

PR #218, opened 2024-12-16
RSCBC-66: Implement collections management in core

chvck:


sdk/couchbase-core/src/cbconfig/mod.rs
```diff
@@ -136,3 +136,36 @@ pub struct TerseConfig {
     #[serde(alias = "clusterCapabilities")]
     pub cluster_capabilities: HashMap<String, Vec<String>>,
 }
+
+#[derive(Deserialize, Debug, Clone)]
+#[non_exhaustive]
+pub struct CollectionManifestCollection {
+    #[serde(rename = "uid")]
```
Westwooo, line 143: Do we need this for uid, name and history?

Westwooo (approved):

sdk/couchbase-core/src/cbconfig/mod.rs
chvck: Probably not, but it doesn't hurt to be explicit

PR #220, opened 2024-12-18
RSCBC-54: Implement Data Structures

Matt-Woz:


sdk/couchbase/src/collection_ds.rs
```diff
@@ -0,0 +1,417 @@
...
+use std::collections::HashMap;
+
+#[derive(Clone)]
+pub struct CouchbaseList<'a> {
+    pub collection: &'a Collection,
+    pub id: String,
+}
+
+impl Collection {
+    pub fn list(&self, id: impl Into<String>) -> CouchbaseList {
```
chvck, line 16: It'd be a good idea for all of the constructor type functions to accept options blocks so that we can add functionality like tracing in the future.

Matt-Woz: Done

sdk/couchbase/src/collection_ds.rs
```diff
@@ -0,0 +1,444 @@
...
+                &self.id,
+                &[MutateInSpec::array_add_unique("", value, None)?],
+                MutateInOptions::builder()
+                    .store_semantics(StoreSemantics::Upsert)
+                    .build(),
+            )
+            .await;
+
+        if let Err(e) = res {
+            return if e.msg.contains("subdoc path exists") {
```
chvck, line 279: Do we not have a way to detect these errors without string parsing? If not we should probably add a TODO here to fix this when we add that.

Matt-Woz: Done

chvck (approved):

PR #221, opened 2024-12-19
RSCBC-14: Use structured logging in core

chvck:


sdk/couchbase-core/src/memdx/client.rs
```diff
@@ -147,8 +148,8 @@ impl Client {
                             match input {
                                 Ok(mut packet) => {
                                     trace!(
-                                        "Resolving response on {}. Opcode={}. Opaque={}. Status={}",
-                                        opts.client_id,
+                                        context=&opts.logger_context;
+                                        "Resolving response. Opcode={}. Opaque={}. Status={}.",
```
Matt-Woz, line 152: Minor, but why not use the same structured field style here as elsewhere? e.g.

```
trace!(
    context = &opts.logger_context,
    opcode = packet.op_code,
    opaque = packet.opaque,
    status = packet.status;
    "Resolving response"
);
```

sdk/couchbase-core/src/memdx/client.rs
```diff
@@ -319,8 +326,8 @@ impl Dispatcher for Client {
         let op_code = packet.op_code;
 
         trace!(
-            "Writing request on {}. Opcode={}. Opaque={}",
-            &self.client_id,
+            context=&self.log_context;
+            "Writing request. Opcode={}. Opaque={}.",
```
Matt-Woz, line 330: Same as above

sdk/couchbase-core/src/memdx/op_bootstrap.rs
```diff
@@ -103,7 +103,7 @@ impl OpBootstrap {
             {
                 Ok(r) => Some(r),
                 Err(e) => {
-                    warn!("Get error map failed {}", e);
+                    warn!(context=dispatcher.log_context(), e:err; "Get error map failed {} {}", dispatcher.log_context(), e);
```
Matt-Woz, line 106: I think the context and error will get logged twice here? (And the Auth Failed & Select bucket below)

sdk/couchbase-core/src/memdx/client.rs
chvck: I just missed it, which is a flaw in this approach - there aren't really any guardrails to it. Unfortunately I'm unsure there's a better approach.

Matt-Woz (approved):

PR #226, opened 2025-01-16
Refactor query/analytics/search params/raw to remove value from API

chvck:


sdk/couchbase/src/options/query_options.rs
```diff
@@ -118,9 +119,25 @@ impl QueryOptions {
...
-        self.named_parameters = Some(named_parameters);
-        self
+    pub fn add_named_parameter<T: Serialize>(
+        mut self,
+        key: impl Into<String>,
+        value: T,
+    ) -> error::Result<Self> {
+        let value = serde_json::to_value(&value)?;
+
+        match self.named_parameters {
```
Westwooo, line 129: nit: could do: 

```
if let Some(mut params) = self.named_params {
...
} else {
...
}
```

sdk/couchbase/src/options/query_options.rs
```diff
@@ -133,9 +150,19 @@ impl QueryOptions {
...
     }
 
-    pub fn positional_parameters(mut self, positional_parameters: Vec<Value>) -> Self {
-        self.positional_parameters = Some(positional_parameters);
-        self
+    pub fn add_positional_parameter<T: Serialize>(mut self, parameters: T) -> error::Result<Self> {
+        let parameters = serde_json::to_value(&parameters)?;
+
+        match self.positional_parameters {
+            Some(mut params) => {
```
Westwooo, line 157: Same nit here and in add_raw.

Westwooo (approved):
LGTM, just one nit

sdk/couchbase/src/options/query_options.rs
chvck: Interestingly clippy isn't highlighting that as a recommendation

PR #230, opened 2025-01-20
RSCBC32: Refactor memdx packet to use borrows

chvck:


sdk/couchbase-core/src/memdx/ops_crud.rs
```diff
@@ -52,12 +52,17 @@ impl OpsCrud {
...
         let framing_extras = if !ext_frame_buf.is_empty() {
-            Some(ext_frame_buf)
+            Some(ext_frame_buf.as_slice())
         } else {
             None
         };
 
-        let key = self.encode_collection_and_key(request.collection_id, request.key)?;
+        let buf = &mut [0; 255];
+        let key = if OpCode::supports_collections(OpCode::Set) {
```
Matt-Woz, line 61: What's the point of these checks, won't this always be true?

Matt-Woz (approved):

PR #231, opened 2025-01-20
RSCBC-75: Rework memdx request persistency through pending op

chvck:


Westwooo (approved):
LGTM just one question

PR #232, opened 2025-01-20
RSCBC-67: Move to rust v1.85-beta

chvck:


sdk/couchbase-core/src/agent_ops.rs
```diff
@@ -32,70 +32,67 @@ impl Agent {
         self.inner.bucket_features().await
     }
 
-    pub async fn upsert<'a>(&self, opts: UpsertOptions<'a>) -> Result<UpsertResult> {
+    pub async fn upsert(&self, opts: UpsertOptions<'_>) -> Result<UpsertResult> {
```
Westwooo, line 35: Has there been a change to the lifetime notation? I'm guessing that this means the same as before?

sdk/couchbase-core/tests/agent_ops.rs
```diff
@@ -659,13 +659,13 @@ async fn upsert_allocations() {
...
 
-    let expected_allocs = if test_is_ssl().await { 23 } else { 21 };
-    dhat::assert!(
-        total_allocs <= 23,
-        "Expected max {} allocations, was {}",
-        expected_allocs,
-        total_allocs
-    );
+    let expected_allocs = if test_is_ssl().await { 19 } else { 21 };
+    // dhat::assert!(
```
Westwooo, line 663: Is this supposed to be commented out?

sdk/couchbase-core/src/crudcomponent.rs
```diff
@@ -691,16 +686,9 @@ impl<
...
                                 async |client: Arc<KvClientManagerClientType<M>>| {
-                                    operation(
-                                        collection_id,
-                                        manifest_rev,
-                                        endpoint.clone(),
-                                        vb_id,
-                                        client,
-                                    )
-                                    .await
+                                    operation(collection_id, manifest_rev, vb_id, client).await
```
Westwooo, line 691: Why doesn't this need endpoint any more?

Westwooo (dismissed):
LGTM, just a couple questions

sdk/couchbase-core/tests/agent_ops.rs
chvck: No :/

sdk/couchbase-core/src/crudcomponent.rs
chvck: It was never being used, and we can probably get it from client itself if we need it.

Westwooo (approved):

PR #234, opened 2025-01-22
RSCBC-3: Implement query index management in the core

Matt-Woz:


sdk/couchbase-core/src/queryx/index.rs
```diff
@@ -0,0 +1,61 @@
...
+            IndexState::Online => "Online",
+            IndexState::Offline => "Offline",
+            IndexState::Abridged => "Abridged",
+            IndexState::Scheduled => "Scheduled",
+        };
+        write!(f, "{}", state_str)
+    }
+}
+
+impl From<IndexState> for String {
```
chvck, line 48: I don't think we need this given we have Display do we?

sdk/couchbase-core/src/queryx/query.rs
```diff
@@ -138,4 +147,538 @@ impl<C: Client> Query<C> {
...
+                let encoded_bucket = encode_value(&opts.bucket_name)?;
+                format!(
+                    "(keyspace_id={} AND bucket_id IS MISSING) OR bucket_id={}",
+                    encoded_bucket, encoded_bucket
+                )
+            } else {
+                "1=1".to_string()
+            }
+        } else if opts.collection_name.is_some() && opts.scope_name.is_some() {
+            let scope_name = normalise_default_name(opts.scope_name.as_ref().unwrap());
```
chvck, line 163: I feel like there must be a more idiomatic way of doing this than checking is_some and then unwrapping.

sdk/couchbase-core/src/queryx/query.rs
```diff
@@ -138,4 +147,538 @@ impl<C: Client> Query<C> {
...
+        }
+
+        let query_opts = QueryOptions::new()
+            .statement(qs)
+            .on_behalf_of(opts.on_behalf_of.clone());
+
+        let mut res = self.query(&query_opts).await;
+
+        // This is a bit mad...
+        match res {
```
chvck, line 271: We should check with PM which server versions we need to support, statements with IF NOT EXISTS are supported as of 7.1 and 7.0 is already end of maintenance and EOL in July - so is 7.1 actually.

sdk/couchbase-core/src/queryx/query.rs
```diff
@@ -138,4 +147,538 @@ impl<C: Client> Query<C> {
...
+            },
+            Ok(_) => Ok(()),
+        }
+    }
+
+    pub async fn build_deferred_indexes(
+        &self,
+        opts: &BuildDeferredIndexesOptions,
+    ) -> error::Result<()> {
+        let opts = GetAllIndexesOptions {
```
chvck, line 470: This makes me think that at this submodule level the fields on these options should be borrows. What do you think?

sdk/couchbase-core/src/queryx/query.rs
```diff
@@ -138,4 +147,538 @@ impl<C: Client> Query<C> {
...
+            on_behalf_of: opts.on_behalf_of.clone(),
+        };
+
+        let indexes = self.get_all_indexes(&opts).await?;
+
+        let mut deferred_indexes: HashMap<String, Vec<DeferredIndexName>> = HashMap::new();
+        for index in indexes {
+            if index.state == IndexState::Deferred {
+                let (bucket, scope, collection) = index_to_namespace_parts(&index);
+                let deferred_index = DeferredIndexName {
```
chvck, line 483: This could potentially be expensive too as we're allocating every string.

sdk/couchbase-core/src/queryx/query_options.rs
```diff
@@ -523,3 +523,344 @@ impl Serialize for QueryOptions {
...
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub struct WatchIndexesOptions {
+    pub(crate) bucket_name: String,
+    pub(crate) scope_name: Option<String>,
+    pub(crate) collection_name: Option<String>,
+    pub(crate) timeout: Duration,
```
chvck, line 821: I think this timeout probably shouldn't exist. The watch operation will just run until the future is dropped.

sdk/couchbase-core/src/queryx/index.rs
chvck: It still seems like Display should cover this to me though. Is there a reason that we want to keep them separated?

sdk/couchbase-core/src/queryx/query.rs
chvck: Nice!

sdk/couchbase-core/src/queryx/index.rs
Matt-Woz: Ah I didn't realise that fmt::Display allows us to use `to_string()` (I was using .into())

chvck (approved):

PR #235, opened 2025-01-23
RSCBC-19: Rework the memdx error model

chvck:


sdk/couchbase-core/src/memdx/error.rs
```diff
@@ -2,168 +2,165 @@ use serde::Deserialize;
...
-            _ => false,
-        }
+    pub fn is_dispatch_error(&self) -> bool {
+        matches!(self.inner.kind, ErrorKind::Dispatch { .. })
     }
 
-    pub fn is_unknown_collection_id_error(&self) -> bool {
-        match self.kind.as_ref() {
-            ErrorKind::Server(e) => e.kind == ServerErrorKind::UnknownCollectionID,
+    pub fn is_server_error_kind(&self, kind: ServerErrorKind) -> bool {
```
Matt-Woz, line 122: Nice, I like the emphasis on the helper methods, matching can get unhandily very quickly especially w/ in the higher level errors, and types like this which are shared between multiple kinds

sdk/couchbase-core/src/memdx/error.rs
```diff
@@ -181,25 +178,19 @@ pub enum ErrorKind {
...
-        arg: String,
-    },
-    NoSupportedAuthMechanisms,
-    #[non_exhaustive]
-    Json {
-        msg: String,
+        arg: Option<String>,
     },
+    Decompression,
+    Message(String),
```
Matt-Woz, line 193: I presume this is just for anything that doesn't fit into the other categories?

Matt-Woz (approved):

sdk/couchbase-core/src/memdx/error.rs
chvck: Yeah, it's basically just "CouchbaseException"

PR #245, opened 2025-02-26
RSCBC-12: Implement tracing

Matt-Woz:


sdk/couchbase/src/threshold_log_tracer.rs
```diff
@@ -0,0 +1,502 @@
...
+                        "top_requests": top_requests,
+                    }),
+                )
+            })
+            .collect();
+
+        if !log_output.is_empty() {
+            match serde_json::to_string(&log_output) {
+                Ok(log_output_str) => {
+                    tracing::warn!("Operations over threshold: {}", log_output_str)
```
Matt-Woz, line 299: I think given we're using tracing for spans already, it probably makes sense to convert all of our logging to use `tracing` instead, since tracing can be configured to also emit `log` records. See the [tracing docs](https://docs.rs/tracing/latest/tracing/#log-compatibility) on log crate compatibility,

sdk/couchbase-core/src/agent.rs
```diff
@@ -462,6 +489,7 @@ impl Agent {
                         info!("Orphan : {:?}", packet);
                     }),
                     disable_decompression: opts.compression_config.disable_decompression,
+                    tracing: tracing.clone(),
```
Matt-Woz, line 492: I'm not sure this is the best way to get the tracing component down the KV stack to the KvClient, but I can't see a nicer way -- Given spans can be manipulated w/ static functions, I did also experiment with just passing the TracingConfig (i.e. ClusterLabels) in with the KvConfig, but I don't think it was much better

PR #249, opened 2025-03-19
Update public durability level to use newtype pattern

chvck:


sdk/couchbase/src/durability_level.rs
```diff
@@ -1,49 +1,62 @@
...
+    PersistToMajority,
+    Other(String),
+}
 
-    fn try_from(value: String) -> Result<Self, Self::Error> {
-        Self::try_from(value.as_str())
+pub(crate) fn parse_optional_durability_level_to_memdx(
+    durability_level: Option<DurabilityLevel>,
+) -> Option<couchbase_core::memdx::durability_level::DurabilityLevel> {
+    if let Some(durability_level) = durability_level {
```
Peter-Searby, line 46: I feel like this reads better as just a match, or maybe [`.and_then`](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then), rather than
```rust
if let .. {
  // explicit return
}
// implicit return
```

sdk/couchbase/src/durability_level.rs
```diff
@@ -1,49 +1,62 @@
...
+
+    pub const MAJORITY: DurabilityLevel = DurabilityLevel(InnerDurabilityLevel::Majority);
+
+    pub const MAJORITY_AND_PERSIST_ACTIVE: DurabilityLevel =
+        DurabilityLevel(InnerDurabilityLevel::MajorityAndPersistActive);
+
+    pub const PERSIST_TO_MAJORITY: DurabilityLevel =
+        DurabilityLevel(InnerDurabilityLevel::PersistToMajority);
+
+    pub(crate) fn other(val: String) -> DurabilityLevel {
```
Peter-Searby, line 17: Perhaps I'm just missing it, but I don't see where this is used. What is it for?

Westwooo (dismissed):
lgtm

sdk/couchbase/src/durability_level.rs
chvck: This covers the case where the mgmt API returns a durability level that we don't know about. The choice is this or to return an error. It's not used "right now" but will be shortly when i add the bucket manager to the public API.

Peter-Searby: Makes sense, thanks!

sdk/couchbase/src/durability_level.rs
chvck: Agree, can just be collapsed into a single match.

sdk/couchbase/src/durability_level.rs
```diff
@@ -1,49 +1,58 @@
...
-    fn try_from(value: String) -> Result<Self, Self::Error> {
-        Self::try_from(value.as_str())
+pub(crate) fn parse_optional_durability_level_to_memdx(
+    durability_level: Option<DurabilityLevel>,
+) -> Option<couchbase_core::memdx::durability_level::DurabilityLevel> {
+    match durability_level {
+        Some(DurabilityLevel(InnerDurabilityLevel::Majority)) => {
+            Some(couchbase_core::memdx::durability_level::DurabilityLevel::MAJORITY)
+        }
+        Some(DurabilityLevel(InnerDurabilityLevel::MajorityAndPersistActive)) => Some(
```
Peter-Searby, line 50: nit, but having `{...}` around the other clauses and not this one is a bit inconsistent

chvck: Interestingly my IDE automatically reformats to this, a rustfmt thing maybe?

Peter-Searby: Very odd, does it perhaps let you format them all like this one instead?
Either way, since it's a nit, I'll approve anyway

Peter-Searby (approved):

PR #251, opened 2025-03-19
RSCBC-10: Add support for bucket management to public API

chvck:


sdk/couchbase/tests/bucket_management.rs
```diff
@@ -0,0 +1,287 @@
...
+        )
+        .await;
+
+        manager.delete_bucket(&bucket_name, None).await.unwrap();
+        verify_bucket_deleted(&manager, &bucket_name).await;
+    })
+}
+
+#[test]
+fn test_delete_bucket() {
```
Westwooo, line 81: Do we need this since we are deleting and verifying in the earlier tests?

sdk/couchbase/tests/bucket_management.rs
```diff
@@ -0,0 +1,287 @@
...
+
+        verify_bucket_created(&manager, &bucket_name).await;
+
+        let updated_settings = settings.ram_quota_mb(200);
+        manager.update_bucket(updated_settings, None).await.unwrap();
+
+        try_until(
+            tokio::time::Instant::now() + Duration::from_secs(30),
+            Duration::from_millis(100),
+            "Bucket was not created in time",
```
Westwooo, line 62:
```suggestion
            "Bucket was not updated in time",
```

Westwooo (approved):
lgtm just a couple questions and nits

sdk/couchbase-core/src/mgmtx/bucket_settings.rs
```diff
@@ -191,7 +191,7 @@ impl From<BucketSettingsJson> for BucketDef {
                 mutable_bucket_settings: MutableBucketSettings {
                     flush_enabled: settings.controllers.as_ref().map(|c| {
                         if let Some(f) = &c.flush {
-                            f == "enabled"
+                            !f.is_empty()
```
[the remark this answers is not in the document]
chvck: It was wrong, if the flush field is populated then it's enabled. It doesn't get populated with "enabled" either.

sdk/couchbase/tests/bucket_management.rs
chvck: I think it's just nice to have, it can be run standalone to reproduce any delete issues.

sdk/couchbase/src/bucket_settings.rs
```diff
@@ -0,0 +1,478 @@
...
+#[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub(crate) enum InnerCompressionMode {
+    Off,
+    Passive,
+    Active,
+    Unknown(String),
+}
+
+#[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct ConflictResolutionType(InnerConflictResolutionType);
```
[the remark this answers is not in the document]
chvck: It allows us to handle unknown values without some oddities that occur when trying to use an enum for this.

sdk/couchbase/src/clients/bucket_mgmt_client.rs
```diff
@@ -0,0 +1,423 @@
...
+
+        if let Some(history_retention_bytes) = settings.history_retention_bytes {
+            core_settings = core_settings.history_retention_bytes(history_retention_bytes);
+        }
+
+        if let Some(history_retention_seconds) = settings.history_retention_seconds {
+            core_settings = core_settings.history_retention_seconds(history_retention_seconds);
+        }
+
+        if settings.conflict_resolution_type.is_some() {
```
Matt-Woz, line 297: Same here and below with the validation

sdk/couchbase/src/bucket_settings.rs
```diff
@@ -0,0 +1,478 @@
...
+
+impl From<couchbase_core::mgmtx::bucket_settings::BucketDef> for BucketSettings {
+    fn from(value: BucketDef) -> Self {
+        Self {
+            name: value.name,
+            ram_quota_mb: value
+                .bucket_settings
+                .mutable_bucket_settings
+                .ram_quota_mb
+                .unwrap_or(0),
```
Peter-Searby, line 438: Why is a default of 0 being used? We don't seem to do something like that for any other settings?

sdk/couchbase/src/clients/bucket_mgmt_client.rs
Peter-Searby: I also agree here. An example of why it's better to leave this kind of validation to ns_server is the storageBackend check. In 7.6, we added the capability to migrate the storageBackend (https://docs.couchbase.com/server/current/manage/manage-buckets/migrate-bucket.html#perform_migration), which this code doesn't permit.
If we want to explicitly disable something in the SDK, then that would be reasonable, but reproducing the ns_server validation just makes inconsistency likely.

sdk/couchbase/src/bucket_settings.rs
chvck: As far as I'm aware this can't actually be empty, but we use the same type for create and get so it has to be optional at the core level but we don't allow it to be optional in the public API.

chvck: That's right, for create or update it just leaves it up to server to figure out what it wants to do.

sdk/couchbase/src/bucket_settings.rs
```diff
@@ -0,0 +1,478 @@
...
+        self.bucket_type = Some(bucket_type.into());
+        self
+    }
+
+    pub fn storage_backend(mut self, storage_backend: impl Into<StorageBackend>) -> Self {
+        self.storage_backend = Some(storage_backend.into());
+        self
+    }
+
+    pub(crate) fn validate(&self) -> error::Result<()> {
```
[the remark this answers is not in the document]
chvck: The other downside is that it might make it more difficult to raise errors like that as a specific invalid argument error as I don't believe that the server replies with any specific error codes?

sdk/couchbase/src/bucket_manager.rs
```diff
@@ -0,0 +1,89 @@
...
+            .get_bucket(
+                bucket_name.into(),
+                opts.into().unwrap_or(GetBucketOptions::default()),
+            )
+            .await
+    }
+
+    pub async fn create_bucket(
+        &self,
+        settings: BucketSettings,
```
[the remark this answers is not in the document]
chvck: Not on purpose but agree.

sdk/couchbase/src/bucket_settings.rs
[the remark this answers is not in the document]
chvck: We'd need an error code to indicate what sort of error it is really, in order to know how to translate the error in the SDKs.

Matt-Woz (approved):
LGTM 👍

PR #258, opened 2025-04-01
RSCBC-71: Add support for user management to core

chvck:


.github/workflows/tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - 7.1.1
           - 7.0.3
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
Peter-Searby, line 23: Why is this being changed?

sdk/couchbase-core/src/mgmtoptions.rs
```diff
@@ -579,3 +580,426 @@ impl<'a> EnsureBucketOptions<'a> {
         self
     }
 }
+
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct GetUserOptions<'a> {
+    pub username: &'a str,
+    pub domain_name: &'a str,
```
Peter-Searby, line 588: `domain_name` seems like the wrong name for this. Shouldn't it just be `domain` or `auth_domain` (for consistency with rfc)? Otherwise it sounds like the internet domain kind of domain

sdk/couchbase-core/src/mgmtx/mgmt.rs
```diff
@@ -221,369 +216,6 @@ impl<C: Client> Management<C> {
 
         parse_response_json(resp).await
     }
-
```
Peter-Searby, line 224: Why are we deleting a bunch of other API code? If it's refactoring which helps this commit, then it would be a lot easier to review as two separate commits

sdk/couchbase-core/src/mgmtx/user_json.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct GroupJson {
+    pub id: String,
+    pub description: Option<String>,
+    pub roles: Vec<RoleJson>,
+    #[serde(rename = "ldap_group_ref", skip_serializing_if = "Option::is_none")]
+    pub ldap_group_reference: Option<String>,
+}
+
+#[derive(Deserialize, Debug, Clone, PartialOrd, PartialEq)]
+pub struct UserJson {
```
Peter-Searby, line 55: How does this handle new fields added by ns_server? Just wondering because there's at least three new fields which will be coming in 8.0

.github/workflows/tests.yml
chvck: 20.04 is being deprecated and they're running brown outs which results in test failures.

sdk/couchbase-core/src/mgmtx/mgmt.rs
Peter-Searby: Ah, that's ok actually, makes sense

sdk/couchbase-core/src/mgmtx/user_helper.rs
```diff
@@ -0,0 +1,99 @@
...
+    on_behalf_of_info: Option<&'a OnBehalfOfInfo>,
+
+    confirmed_endpoints: Vec<&'a str>,
+}
+
+impl<'a> EnsureUserHelper<'a> {
+    pub fn new(
+        user_agent: &'a str,
+        username: &'a str,
+        domain_name: &'a str,
```
Peter-Searby, line 25: I assume this was just missed?

chvck: Yeah, totally forgot it.

Peter-Searby (dismissed):

sdk/couchbase-core/tests/users.rs
```diff
@@ -0,0 +1,305 @@
...
+    assert_eq!(actual.user.roles, expected.roles);
+}
+
+async fn create_and_ensure_user(agent: &Agent, user: &User) {
+    agent
+        .upsert_user(&UpsertUserOptions::new(user, "local"))
+        .await
+        .unwrap();
+
+    loop {
```
Peter-Searby, line 217: It seems like there's no way of leaving this loop when things go wrong? I assume there's a higher level timeout that would catch it, but I'm just wondering if that timeout would take a while to be hit, which could be a nuisance when these tests start failing?

chvck: Yeah fair point. A general thing we have throughout our tests is a lack of timeouts at present. I'll update here and later do a thorough sweep to apply timeouts to operations.

Peter-Searby (approved):

PR #261, opened 2025-05-01
RSCBC-19: Implement frontend error model

chvck:


sdk/couchbase-core/src/crudcomponent.rs
```diff
@@ -616,8 +745,40 @@ impl<
...
+                            .set_scope_name(opts.scope_name.to_string());
+
+                        Error::new_contextual_memdx_error(e)
+                    })
+                    .map_ok(|resp: LookupInResponse| LookupInResult {
+                        value: resp
+                            .ops
+                            .into_iter()
+                            .map(|o| {
+                                let err = if let Some(e) = o.err {
```
Matt-Woz, line 762: Think we can use .map for this?

sdk/couchbase-core/src/crudcomponent.rs
```diff
@@ -648,15 +809,49 @@ impl<
...
                             seqno: t.seqno,
                         });
 
                         MutateInResult {
-                            value: resp.ops,
+                            value: resp
+                                .ops
+                                .into_iter()
+                                .map(|o| {
+                                    let err = if let Some(e) = o.err {
```
Matt-Woz, line 833: Same here with .map

sdk/couchbase/src/collection_ds.rs
```diff
@@ -305,20 +303,19 @@ impl CouchbaseSet<'_> {
...
                     .await;
                 if let Err(e) = res {
-                    // TODO Check for errors properly once we have proper error types
-                    if e.msg.contains("CAS mismatch") || e.msg.contains("Key exists") {
-                        continue;
-                    } else {
-                        return Err(e);
-                    }
+                    match e.kind() {
+                        crate::error::ErrorKind::DocumentExists => continue,
```
Matt-Woz, line 307: NIT: Would it be nicer to use `A | B => continue` for common case behaviour?

Matt-Woz (commented):
Mostly LGTM, just a few comments

sdk/couchbase/src/collection_ds.rs
chvck: Personally I prefer this format as I find it more readable, especially when more cases are added.

Matt-Woz: Yeah fair enough, I suppose it doesn't make too big a difference when the expression is just one word

Matt-Woz (approved):

PR #263, opened 2025-05-13
RSCBC-53: Add support for kv error map

chvck:


sdk/couchbase-core/src/errmapcomponent.rs
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+    }
+
+    pub fn should_retry(&self, status: &Status) -> bool {
+        let err_map = self.err_map.load();
+        if let Some(err_map) = err_map.as_ref() {
+            if let Some(err_data) = err_map.errors.get(&status.into()) {
+                for attr in &err_data.attributes {
+                    if attr == "retry-now" || attr == "retry-later" || attr == "auto-retry" {
```
Matt-Woz, line 71: Does this `auto-retry` attribute actually exist? I see a bunch of SDKs handling it but I can't see any references to it in the KV docs or usage in the [error map](https://github.com/couchbase/kv_engine/blob/master/etc/couchbase/kv/error_maps/error_map_v2.json)

chvck: Unclear, but for consistency I added it

Matt-Woz (approved):

PR #265, opened 2025-05-21
RSCBC-34: Implement more retry reasons

chvck:


sdk/couchbase-core/src/kvclientmanager.rs
```diff
@@ -267,29 +267,26 @@ where
...
-                        // This was a dispatch error, so we can just try with
-                        // a different client instead...
-                        // TODO: Log something
-                        manager
-                            .shutdown_client(Some(endpoint), client)
-                            .await
-                            .unwrap_or_default();
-                        continue;
-                    }
+    let client = manager.get_client(endpoint).await?;
```
Matt-Woz, line 270: Why are we not iterating through the clients here anymore?

chvck: If dispatch failed previously then we'd instantly retry. Now we bubble the error up and let the retry orchestration decide when and whether to retry the request.

Matt-Woz (approved):

PR #270, opened 2025-05-27
Improve query test reliability

chvck:


sdk/couchbase-core/src/httpcomponent.rs
```diff
@@ -234,7 +234,7 @@ impl<C: Client> HttpComponent<C> {
         Fut: Future<Output = error::Result<bool>> + Send,
         T: NodeTarget,
     {
-        let mut attempt_idx = 0;
+        let mut attempt_idx = 1;
```
DemetrisChr, line 237: What's the reason for changing the initial value to 1? 

The calculation in the backoff calculator is `min * factor ^ attempt_idx`. Shouldn't the first backoff be `min` which requires the initial `attempt_idx` be 0?

DemetrisChr (approved):

PR #272, opened 2025-06-05
RSCBC-28: Add support for cluster config push

chvck:


sdk/couchbase-core/src/agent.rs
```diff
@@ -126,14 +130,38 @@ struct AgentComponentConfigs {
 }
 
 impl AgentInner {
-    pub async fn apply_config(&self, config: ParsedConfig) {
-        let mut state = self.state.lock().await;
+    pub async fn unsolicited_packet_handler(&self, packet: ResponsePacket) {
+        if packet.op_code == OpCode::Set {
```
Matt-Woz, line 134: Why do we have specific behaviour for Set?

sdk/couchbase-core/src/configmanager.rs
```diff
@@ -0,0 +1,289 @@
...
+        self.watcher.watch(self.watcher_shutdown_tx.subscribe())
+    }
+
+    pub fn reconfigure(&self, config: ConfigManagerMemdConfig) -> crate::error::Result<()> {
+        self.watcher.reconfigure(ConfigWatcherMemdConfig {
+            endpoints: config.endpoints,
+        })
+    }
+
+    async fn perform_out_of_bound_fetch(
```
Matt-Woz: is `out_of_bound` a typo?

sdk/couchbase-core/src/agent.rs
chvck: Set just corresponds to the op code that the server are using when sending config push.

sdk/couchbase-core/src/configmanager.rs
chvck: Yup

Matt-Woz (approved):

PR #277, opened 2025-07-08
Move core options into their own module

chvck:


DemetrisChr (dismissed):
I'm not familiar with the conventions in Rust, so feel free to ignore this, but would it be better if the namespaces were e.g. `couchbase_core::options::agent` and `couchbase_core::options::query` instead of `couchbase_core::options::agentoptions` and `couchbase_core::options::queryoptions`?

Since they are already in the `options` module, suffixing them with options might be superfluous.

PR #279, opened 2025-07-16
RSCBC-41: Add support for waituntilready to core

chvck:


sdk/couchbase-core/src/diagnosticscomponent.rs
```diff
@@ -55,47 +70,65 @@ pub struct DiagnosticsComponent<C: Client, M: KvClientManager, CM: ConfigManager
...
-            }),
+
+            state,
         }
     }
 
     pub fn reconfigure(&self, config: DiagnosticsComponentConfig) {
-        let mut state_guard = self.state.lock().unwrap();
-        state_guard.bucket = config.bucket;
+        let mut state = self.state.lock().unwrap();
```
Peter-Searby, line 109: Why was state_guard renamed to state?
Isn't it better to keep guard in the name, so that it's obvious that there's a lock being held here

sdk/couchbase-core/src/kvclientpool.rs
```diff
@@ -307,6 +324,14 @@ where
         let notify = self.shutdown_notify.clone();
 
         tokio::spawn(async move {
+            {
+                let guard = state.lock().await;
+                if let Some(e) = &guard.connection_error {
+                    // TODO: Make configurable.
+                    sleep(Duration::from_millis(5000)).await;
```
Peter-Searby, line 331: This sleep on error logic is a bit curious. Is this from an RFC, or is it just an ad-hoc approach? I'd be interested to know why this sleep is being used, and why some form of polling isn't more appropriate?

Also, is there a ticket for the TODO above? If not, then it's probably just going to get forgotten, so I don't really see the point in the comment, if there's no intent to actually do it.

chvck: I'll update the TODO with a reference to the relevant ticket. Interestingly I took the value from Go, but have no idea where it came from - it's not in the RFC... Will follow up with SDK team.

sdk/couchbase-core/src/diagnosticscomponent.rs
chvck: Hmm maybe, I just felt it less verbose. The type itself is `MutexGuard` anyway, but maybe you're correct.

sdk/couchbase-core/src/kvclientpool.rs
chvck: There appears to not be cross-SDK consistency on this, so I think we just go with it for the time being. It appears that a wait between attempts is applied in at least Dotnet too.

Peter-Searby: Well that's unfortunate, but ok

Peter-Searby (approved):

PR #285, opened 2025-07-23
RSCBC-79: Revisit all enums

chvck:
Motivation
----------
We need to ensure that we're using newtype pattern for enums that the server really owns. We also need to ensure that we can further expand enums that we may need to in the future.

### DIFF
--- a/sdk/couchbase-core/src/memdx/durability_level.rs
+++ b/sdk/couchbase-core/src/memdx/durability_level.rs
@@ -2,6 +2,7 @@ use std::fmt::Display;
 use std::time::Duration;
 
 #[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+#[non_exhaustive]
 pub struct DurabilityLevel(InnerDurabilityLevel);
 
 impl DurabilityLevel {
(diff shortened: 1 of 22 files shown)


sdk/couchbase-core/src/memdx/durability_level.rs
```diff
@@ -2,6 +2,7 @@ use std::fmt::Display;
 use std::time::Duration;
 
 #[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+#[non_exhaustive]
```
Matt-Woz, line 5: What does `#[non_exhaustive]` actually do in this/similar cases? I thought that since `InnerDurabilityLevel` isn't public, downstream users wouldn't be able to create the struct directly anyway?

chvck: Possibly it doesn't much, but only because InnerDurabilityLevel has the Other variant on it. The intention here is less about users creating the struct directly and preventing them from performing pattern matching on it without a wildcard branch - can easily see a case where users are doing pattern matching to convert from their own types into ours.

Matt-Woz: Good point, though, I think it wouldn't let you compile without a wildcard branch regardless since we don't expose the `Other`variant directly. But, I think it still makes sense to add at least for clarity/intention purposes

Matt-Woz (approved):

PR #289, opened 2025-08-05
RSCBC-52: Fill out cluster options

chvck:


sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -47,59 +96,150 @@ impl From<CompressionMode> for couchbase_core::options::agent::CompressionMode {
...
+impl KvOptions {
     pub fn new() -> Self {
-        Self::default()
+        Self {
+            enable_mutation_tokens: None,
+            enable_server_durations: None,
+            num_connections: None,
+            connect_timeout: None,
+            connect_throttle_timeout: None,
+        }
```
Peter-Searby: Isn't this the same as `Self::default()`?

sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -47,59 +96,150 @@ impl From<CompressionMode> for couchbase_core::options::agent::CompressionMode {
...
-        self.kv_connect_timeout = Some(timeout);
+    pub fn disable_mutation_tokens(mut self) -> Self {
+        self.enable_mutation_tokens = Some(false);
         self
     }
+
+    pub fn disable_server_durations(mut self) -> Self {
+        self.enable_server_durations = Some(false);
+        self
+    }
```
Peter-Searby, line 185: This interface feels weird. We default to None for each of these, which I suppose is implicitly `mutation_tokens`/`server_durations` enabled, and we allow disabling, which sets it to `Some(false)`, but we never use the value `Some(true)` as far as I can tell. Perhaps this is just a common pattern?

chvck: I wasn't entirely sure what to do with these tbh. I intend on making all options fields public in another change at which point a user could set this to Some(true). The RFC states that we should have `enable_mutation_tokens: bool`, but making the function `enable` and passing a bool feels weird when it defaults to true anyway. It's possible that this field just shouldn't be optional and be a plain bool - I've just been repeatedly burnt in Go by not knowing if a user actually set an option or not...

sdk/couchbase/src/options/cluster_options.rs
chvck: Will fix, a carry-over from some changes I made where default wasn't present.

sdk/couchbase/src/options/cluster_options.rs
Peter-Searby: Yeah, that makes sense, thanks

Peter-Searby (approved):

PR #292, opened 2025-08-08
RSCBC-97: Support FTS like Pre-Filters while doing Vector Search

willbroadbelt:
Changes:
- Add prefilter optional field to VectorQuery
- Fixed bug with parsing wrong field for checking vectorSearch cluster capability
- Added test for vector search with the prefilter option used


sdk/couchbase-core/src/configparser.rs
```diff
@@ -80,7 +80,7 @@ impl ConfigParser {
         };
 
         let mut features = vec![];
-        if let Some(caps) = config.cluster_capabilities.get("fts") {
+        if let Some(caps) = config.cluster_capabilities.get("search") {
```
chvck, line 83: Good catch.

PR #296, opened 2025-08-12
RSCBC-111: Rework error source fields

chvck:
Motivation
----------
Source is designed to allow users to access underlying platform errors from our error types. In the case of http based errors we don't actually want to do this as it'll leak implementation detail.

Changes
-------
Remove source from error types for services utilising http. Update log messages for memdx errors to include source messages. Update public API error to include any source in the error message.

### DIFF
--- a/sdk/couchbase-core/src/httpx/client.rs
+++ b/sdk/couchbase-core/src/httpx/client.rs
@@ -10,6 +10,7 @@ use base64::Engine;
 use http::header::{CONTENT_TYPE, USER_AGENT};
 use log::trace;
 use reqwest::redirect::Policy;
+use std::error::Error as StdError;
 use std::sync::Arc;
 use std::time::Duration;
 use uuid::Uuid;
@@ -183,12 +184,17 @@ impl Client for ReqwestClient {
                 Response::from(response)
             }),
             Err(err) => {
-                trace!(
+                let mut msg = format!(
                     "Received error on {}. Request id={}. Err: {}",
-                    &self.client_id,
-                    &id,
-                    &err,
+                    &self.client_id, &id, &err,
                 );
+
+                if let Some(source) = err.source() {
+                    msg = format!("{msg}. Source: {source}");
+                }
+
+                trace!("{msg}");
+
                 if err.is_connect() {
                     Err(Error::new_connection_error(err.to_string()))
                 } else {
--- a/sdk/couchbase-core/src/kvclientpool.rs
+++ b/sdk/couchbase-core/src/kvclientpool.rs
@@ -1,5 +1,6 @@
 use std::backtrace::Backtrace;
 use std::collections::HashMap;
+use std::error::Error as StdError;
 use std::future::Future;
 use std::net::SocketAddr;
 use std::ops::{Deref, Sub};
@@ -380,7 +381,6 @@ where
             {
                 let mut guard = state.lock().await;
                 if let Some(e) = &guard.connection_error {
-                    // TODO(RSCBC-52): Make configurable.
                     sleep(throttle_period).await;
                 }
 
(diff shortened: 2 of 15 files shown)


sdk/couchbase-core/src/httpx/client.rs
```diff
@@ -183,12 +184,17 @@ impl Client for ReqwestClient {
...
+                let mut msg = format!(
                     "Received error on {}. Request id={}. Err: {}",
-                    &self.client_id,
-                    &id,
-                    &err,
+                    &self.client_id, &id, &err,
                 );
+
+                if let Some(source) = err.source() {
+                    msg.push_str(&format!(", Source: {source}"));
```
Peter-Searby: It's a bit odd that you used `msg = format!("{msg}...` in one place, and `msg.push_str` here. I don't have a strong preference either way, but it would be good to be consistent, especially within the same pull request

sdk/couchbase-core/src/kvclientpool.rs
```diff
@@ -380,7 +381,6 @@ where
             {
                 let mut guard = state.lock().await;
                 if let Some(e) = &guard.connection_error {
-                    // TODO(RSCBC-52): Make configurable.
```
Peter-Searby, line 383: This pull request doesn't make this configurable, so it shouldn't be removing this comment, right?

sdk/couchbase-core/src/kvclientpool.rs
```diff
@@ -439,7 +439,11 @@ where
                             };
                         }
                         Err(e) => {
-                            debug!("Failed to reconfigure client {}: {}", r.id(), e);
+                            let mut msg = format!("Failed to reconfigure client {}: {}", r.id(), e);
+                            if let Some(source) = e.source() {
+                                msg = format!("{msg} - {source}");
```
Peter-Searby, line 444: See my other comment about consistently formatting these messages

sdk/couchbase-core/src/kvclientpool.rs
```diff
@@ -452,7 +456,12 @@ where
                     };
                 }
                 Err(e) => {
-                    debug!("Error creating new client: {}", &e);
+                    let mut msg = format!("Error creating new client {e}");
+                    if let Some(source) = e.source() {
+                        msg = format!("{msg} - {source}");
```
Peter-Searby, line 461: See my other comment about consistently formatting these messages

sdk/couchbase-core/src/mgmtx/mgmt.rs
```diff
@@ -86,7 +86,7 @@ impl<C: Client> Management<C> {
         self.http_client
             .execute(req)
             .await
-            .map_err(|e| error::Error::new_message_error("could not execute request").with(e))
+            .map_err(|e| error::Error::new_message_error(format!("could not execute request: {e}")))
```
Peter-Searby, line 89: I can see the logic of moving the implementation details out of an error struct, and replacing that with logging, but I don't quite see the point of moving it from a structured form to a merged error message string?

sdk/couchbase/src/error.rs
```diff
@@ -76,6 +76,10 @@ impl Error {
             collection_name,
         );
 
+        if let Some(source) = e.source() {
+            extended_context = extended_context.with_source_message(source.to_string());
```
Peter-Searby, line 80: What's the difference between how source was previously being specified and how it is now being included in extended_context, from a user perspective?

sdk/couchbase-core/src/kvclientpool.rs
chvck: This is already configurable, removing this was just missed at the time.

sdk/couchbase/src/error.rs
chvck: Previously it was totally missing, so the user wouldn't have seen this detail at all.

sdk/couchbase-core/src/mgmtx/mgmt.rs
chvck: Internally to Error all we'd do is make it a merged error message anyway. This just makes intent more clear IMO, I'm not wedded to it though.

sdk/couchbase-core/src/httpx/client.rs
chvck: Fair point, done.

sdk/couchbase-core/src/mgmtx/mgmt.rs
Peter-Searby: Fair enough, if it gets merged anyway then this seems reasonable

Peter-Searby (approved):

PR #298, opened 2025-08-13
Query result metadata should not be async 

Matt-Woz:


chvck (approved):

chvck (changes requested):
Looks good, just needs a RSCBC

PR #302, opened 2025-08-20
RSCBC-118: Rename search index mgmt options

chvck:


sdk/couchbase/src/options/search_index_mgmt_options.rs
```diff
@@ -1,38 +1,38 @@
...
     }
 }
 
 #[derive(Default, Debug, Clone)]
 #[non_exhaustive]
-pub struct DeleteIndexOptions {}
+pub struct DeleteSearchIndexOptions {}
 
-impl DeleteIndexOptions {
+impl DeleteSearchIndexOptions {
```
Matt-Woz: Should this not be `DropSearchIndexOptions`? https://github.com/couchbaselabs/sdk-rfcs/blob/master/rfc/0054-sdk3-management-apis.md?plain=1#L1068

chvck: Good catch.

PR #313, opened 2025-08-28
RSCBC-129: Add action to verify ticket in commit message exists

chvck:


.github/workflows/jira-validator.yml
```diff
@@ -0,0 +1,32 @@
...
+          
+          # Set variables
+          BASE_BRANCH=${{ github.event.pull_request.base.ref }}
+          
+          # Write BASE_BRANCH to GITHUB_ENV so commit checker can use
+          echo "BASE_BRANCH=$BASE_BRANCH" | tee -a $GITHUB_ENV
+
+      - name: Jira Commit Checker
+        id: jira_commit_checker
+        uses: KyMidd/ActionJiraCommitValidate@v1
```
Peter-Searby: This feels a little dangerous (especially with https://arstechnica.com/security/2025/09/software-packages-with-more-than-2-billion-weekly-downloads-hit-in-supply-chain-attack/ fresh in the mind). Given this is not a particularly mature/popular repo, it would be good to tag a specific commit or fork it. We are giving this access to our jira api token, so there is definitely a risk here

chvck: Good point, we'll be using this across several repos so a fork would probably be good.

chvck: Done.

PR #315, opened 2025-08-28
RSCBC-96: Add Orphan Reporter

willbroadbelt:
Motivation
----------
To improve observability of orphaned requests and implement the orphan reporter sections of RFC #67.

Modifications
-------------
- Added OrphanReporterOptions to Cluster options for Orphan reporter configuration
- Created OrphanReporter with buffered aggregation and periodic flush
- Piped OrphanResponseHandler into agents so all orphaned responses are recorded across all connections
- Added unit and integration tests with log capture for validations

Notes
-----
Orphans are only detected for KV currently, so future work will need to be done to make this work with other services.


sdk/couchbase-core/src/memdx/client.rs
```diff
@@ -199,7 +202,15 @@ impl Client {
                                                             match sender.send(Err(Error::new_decompression_error().with(e))).await{
                                                                 Ok(_) => {}
                                                                 Err(e) => {
-                                                                    debug!("Sending response to caller failed: {e}");
+                                                                     debug!("Sending response to caller failed: {e}");
+                                                                     (opts.orphan_handler)(
```
chvck: I don't think that this counts as an orphan

sdk/couchbase-core/src/memdx/client.rs
```diff
@@ -218,17 +229,31 @@ impl Client {
                                             }
                                         }
 
-                                        let resp = ClientResponse::new(packet, context.context.clone());
+                                        let resp = ClientResponse::new(packet.clone(), context.context.clone());
                                         match sender.send(Ok(resp)).await {
                                             Ok(_) => {}
                                             Err(e) => {
                                                 debug!("Sending response to caller failed: {e}");
+                                                (opts.orphan_handler)(
```
chvck: Not an orphan I think, would also mean we don't need to clone the packet above.

sdk/couchbase-core/src/options/agent.rs
```diff
@@ -19,6 +22,24 @@ pub struct AgentOptions {
     pub kv_config: KvConfig,
     pub http_config: HttpConfig,
     pub tcp_keep_alive_time: Option<Duration>,
+    pub orphan_response_handler: OrphanResponseHandler,
+}
+
+impl Debug for AgentOptions {
```
chvck, line 28: We need to be careful doing this. We'll need log redaction implementing in order to log the seed config and bucket name. I'm unsure what the debug impl for authenticator does but it actually probably shouldn't exist - that seems very prone to leaking credentials.

sdk/couchbase-core/src/options/agent.rs
```diff
@@ -19,6 +22,24 @@ pub struct AgentOptions {
     pub kv_config: KvConfig,
     pub http_config: HttpConfig,
     pub tcp_keep_alive_time: Option<Duration>,
+    pub orphan_response_handler: OrphanResponseHandler,
```
chvck: I think this should be optional, not all users will want to use it.

sdk/couchbase-core/src/options/agent.rs
```diff
@@ -262,3 +292,38 @@ impl Default for HttpConfig {
         }
     }
 }
+
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub struct OrphanReporterConfig {
```
chvck: This should probably be in its own file as it doesn't pertain to the agent directly.

sdk/couchbase-core/src/options/agent.rs
```diff
@@ -262,3 +292,38 @@ impl Default for HttpConfig {
...
+    pub enabled: bool,
+    pub reporter_interval: Duration,
+    pub sample_size: usize,
+}
+
+impl OrphanReporterConfig {
+    pub fn enabled(mut self, enabled: bool) -> Self {
```
chvck: Do we need enabled? The existence of an instance of orphan reporter kind of implies that it's enabled.

sdk/couchbase-core/src/options/ondemand_agentmanager.rs
```diff
@@ -20,6 +22,7 @@ pub struct OnDemandAgentManagerOptions {
     pub kv_config: KvConfig,
     pub http_config: HttpConfig,
     pub tcp_keep_alive_time: Option<Duration>,
+    pub orphan_response_handler: OrphanResponseHandler,
```
chvck: Same comment as agent_config

sdk/couchbase-core/src/orphan_reporter.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    fn create_log_object(
+        service: String,
+        heap_items: BinaryHeap<Reverse<OrphanLogItem>>,
+        count: u64,
+    ) -> OrphanLogService {
+        let items: Vec<OrphanLogItem> = heap_items
+            .into_sorted_vec()
+            .into_iter()
```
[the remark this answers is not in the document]
chvck: Any reason why this wasn't applied?

sdk/couchbase-core/tests/orphan_reporter_test.rs
```diff
@@ -0,0 +1,165 @@
...
+            .await
+            .orphan_reporter_handler(orphan_reporter.get_handle());
+
+        let agent = Agent::new(agent_opts).await.unwrap();
+
+        let strat = Arc::new(FailFastRetryStrategy::default());
+        let value = generate_bytes_value(32);
+
+        // Let agent bootstrap
+        tokio::time::sleep(Duration::from_secs(5)).await;
```
chvck: This shouldn't be necessary should it?

sdk/couchbase-core/src/orphan_reporter.rs
```diff
@@ -0,0 +1,220 @@
...
+
+impl OrphanReporter {
+    pub fn new(config: OrphanReporterConfig) -> Self {
+        if !config.enabled {
+            return Self {
+                enabled: config.enabled,
+                sender: None,
+            };
+        }
+        let (tx, mut rx) = mpsc::channel::<OrphanLogItem>(100);
```
chvck: Why 100?

chvck: Why are we using a channel? It seems like we could do all of the checking if we need to add the item to the buffer in the orphan callback itself? Using the channel is going to mean that we allocate every log item regardless of if we then throw it away I think.

sdk/couchbase-core/tests/common/test_config.rs
```diff
@@ -24,6 +25,39 @@ lazy_static! {
     pub static ref LOGGER_INITIATED: AtomicBool = AtomicBool::new(false);
 }
 
+// Shared log capture buffer for log message assertions in core tests
+pub static LOG_CAPTURE: LazyLock<Mutex<String>> = LazyLock::new(|| Mutex::new(String::new()));
+static CAPTURE_LOGS: AtomicBool = AtomicBool::new(false);
+
+fn enable_log_capture() {
```
chvck: Tests within a test file generally run in parallel. If more than 1 test is using log capture then this might do some weird things.

sdk/couchbase-core/src/memdx/client.rs
willbroadbelt: Discussed offline - underlying bug made orphans report here, removed now fixed.

sdk/couchbase-core/src/options/agent.rs
willbroadbelt: Removed it.

sdk/couchbase-core/src/orphan_reporter.rs
willbroadbelt: Fair, changed it to use a Arc<Mutex<State>>> with our orphans + count.

sdk/couchbase-core/src/orphan_reporter.rs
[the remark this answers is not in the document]
willbroadbelt: Its just not valid code, we have to deal with fact the LogItems are wrapped in Reverse for the ordering, so this doesnt work.

sdk/couchbase-core/src/options/agent.rs
willbroadbelt: Ok, Debug was being derived for this but I cant make the type OrphanResponseHandler impl Debug afaict. I can also remove the authenticator field from this debug impl?

sdk/couchbase-core/tests/orphan_reporter_test.rs
willbroadbelt: Replaced with waituntilready, which was needed on my testing env.

sdk/couchbase-core/tests/common/test_config.rs
willbroadbelt: True it could be a problem, I've added a sink/callback that can be used in the reporter, which the tests now use instead to get the output.

sdk/couchbase-core/src/options/agent.rs
chvck: Ah fair point about it already being debug. Probably best to remove authenticator yeah.

sdk/couchbase-core/src/agent.rs
```diff
@@ -541,9 +542,7 @@ impl Agent {
                     connect_timeout,
                     connect_throttle_period: opts.kv_config.connect_throttle_timeout,
                     unsolicited_packet_tx: Some(unsolicited_packet_tx),
-                    orphan_handler: Arc::new(|packet| {
-                        info!("Orphan : {packet:?}");
-                    }),
+                    orphan_handler: opts.orphan_response_handler.clone(),
```
chvck: Why does this need to be cloned?

sdk/couchbase-core/src/orphan_reporter.rs
```diff
@@ -0,0 +1,217 @@
...
+
+            loop {
+                tokio::select! {
+                    _ = tick.tick() => {
+                        let mut st = state_clone.lock().unwrap();
+                        if st.total_count == 0 {
+                            continue;
+                        }
+
+                        let obj = Self::create_log_object("kv".to_string(), mem::take(&mut st.heap), st.total_count);
```
chvck: We might want to pull total_count and heap out of the state before creating the log object, and certainly before logging. As we're using take and total_count is copy anyway we should be able to pull them out and then drop the lock. Locking across the rest of this function is going to be expensive. Something like
```
let (heap, total_count) = {
    let mut st = state_clone.lock().unwrap();
    if st.total_count == 0 {
        continue;
    }

   (mem::take(&mut st.heap), st.total_count)
}
```

chvck (commented):
I think that this is looking good. I'd like to run some benchmarks with orphan reporting enabled vs disabled but that can happen after this goes in.

sdk/couchbase-core/src/orphan_reporter.rs
```diff
@@ -0,0 +1,217 @@
...
+    state: Arc<Mutex<State>>,
+    sample_size: usize,
+    reporter_interval: Duration,
+    log_sink: Option<Arc<OrphanSinkFn>>,
+}
+
+impl OrphanReporter {
+    pub fn new(config: OrphanReporterConfig) -> Self {
+        let state = Arc::new(Mutex::new(State {
+            heap: BinaryHeap::new(),
```
chvck: Suspect this should be with_capacity(sample_size)

willbroadbelt: Done, thanks.

sdk/couchbase-core/src/orphan_reporter.rs
willbroadbelt: Good point, done.

sdk/couchbase-core/src/agent.rs
willbroadbelt: .. it does not, changed, nice catch

sdk/couchbase-core/src/options/agent.rs
willbroadbelt: 👍 removed

sdk/couchbase-core/src/memdx/client.rs
```diff
@@ -226,10 +229,16 @@ impl Client {
...
+                                    } else if let Some(ref orphan_handler) = opts.orphan_handler {
+                                            orphan_handler(
+                                                packet,
+                                                OrphanContext {
+                                                    client_id: opts.client_id.clone(),
+                                                    local_addr: opts.local_addr,
+                                                    peer_addr: opts.peer_addr,
+                                                },
+                                            );
+                                        }
```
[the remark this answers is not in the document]
willbroadbelt: Done

sdk/couchbase-core/src/orphan_reporter.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+struct State {
+    heap: BinaryHeap<Reverse<OrphanLogItem>>,
+    total_count: u64,
+}
+
+// Orphan reporter - Currently only handles/receives KV orphans
+pub struct OrphanReporter {
+    state: Arc<Mutex<State>>,
```
chvck: We might want to make this a RWMutex so that we can take the read lock when recording orphans.

chvck (changes requested):
I think just 1 last comment.

sdk/couchbase-core/src/orphan_reporter.rs
willbroadbelt: Done

sdk/couchbase-core/src/orphan_reporter.rs
```diff
@@ -0,0 +1,240 @@
...
+                .as_deref()
+                .and_then(|f| decode_res_ext_frames(f).ok().flatten())
+                .unwrap_or_default();
+
+            // Read-only
+            let (current_length, current_min) = {
+                let guard = heap.read().unwrap_or_else(|p| {
+                    warn!("OrphanReporter heap poisoned; continuing");
+                    p.into_inner()
+                });
```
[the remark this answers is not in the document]
willbroadbelt: This is best effort metrics logging, panicing isnt appropriate, recovering in this way seems best in the unlikely case it ever happens.

chvck (approved):
Let's do it. Would be quite interested in benchmarks with and without but we can do that later I think.

PR #317, opened 2025-08-29
RSCBC-134: VectorSearch num_candidates should set a default

Matt-Woz:
Motivation
-----------
`k` has to be set in the search payload, and per RFC should default to 3

Changes
---------
Set num_candidates default to 3

### DIFF
--- a/sdk/couchbase-core/src/searchx/query_options.rs
+++ b/sdk/couchbase-core/src/searchx/query_options.rs
@@ -146,8 +146,7 @@ pub struct KnnQuery {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub boost: Option<f32>,
     pub field: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub k: Option<i64>,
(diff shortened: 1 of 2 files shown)


sdk/couchbase-core/src/searchx/query_options.rs
```diff
@@ -161,7 +160,7 @@ impl KnnQuery {
         Self {
             boost: None,
             field: field.into(),
-            k: None,
```
chvck, line 164: I don't think the core should default to anything. I think it should also be skipping serialization if not set. The core is closer to mirroring the rest API than the public API.

chvck (changes requested):

sdk/couchbase-core/src/searchx/query_options.rs
Matt-Woz: True, I changed it to take it in the constructor instead, as it's a required field: https://docs.couchbase.com/server/current/search/search-request-params.html#knn-object

PR #321, opened 2025-09-02
RSCBC-142: Only apply cluster configs that are newer than we have

chvck:


sdk/couchbase-core/src/configmanager.rs
```diff
@@ -243,26 +278,27 @@ impl<M: KvClientManager + 'static> ConfigManagerMemd<M> {
         let (watcher_shutdown_tx, watcher_shutdown_rx) = broadcast::channel(1);
         let bucket = opts.first_config.bucket.clone();
 
+        let (on_new_config_tx, on_new_config_rx) = broadcast::channel::<ParsedConfig>(1);
```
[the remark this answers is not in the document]
chvck: We only need the most recent one.

sdk/couchbase-core/src/configmanager.rs
```diff
@@ -243,26 +278,27 @@ impl<M: KvClientManager + 'static> ConfigManagerMemd<M> {
...
             performing_out_of_band_fetch: AtomicBool::new(false),
 
-            latest_config: Mutex::new(opts.first_config),
+            latest_config: Arc::new(Mutex::new(opts.first_config)),
             latest_version_tx,
             bucket: Mutex::new(bucket),
 
-            on_new_config_txs: Vec::new(),
+            on_new_config_tx,
+            on_new_config_rx,
```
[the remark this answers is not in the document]
chvck: If we drop it then the tx will error.

sdk/couchbase-core/src/configmanager.rs
Peter-Searby: It seems like broadcast channel isn't quite the appropriate model for this, given it leaves us with the unused dummy receiver that we can't drop. Is there a reason we're not using [`sync::watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) for this as well? As long as we used one of the alternative send functions that are able to continue working after all receivers have been dropped, it seems like it should work.

chvck: I think you're correct, it probably should be a watch. The downside to a watch is that it holds a lock for the entire duration of the lifetime of the borrow that it gives out, which also blocks the sender side. Which might be fine.

sdk/couchbase-core/src/agent.rs
```diff
@@ -671,14 +671,13 @@ impl Agent {
...
-                    }
-                    Err(_e) => {
-                        return;
-                    }
-                }
+                let pc = {
+                    // apply_config requires an owned ParsedConfig, as it takes ownership of it.
+                    // Doing the clone within a block also means we can release the lock that
+                    // borrow_and_update() takes as soon as possible.
+                    watch_rx.borrow_and_update().clone()
```
chvck: I don't really love this as it requires the watcher to know about the lock. However, the caller has to know about the channel impl anyway to know what function to call...

sdk/couchbase-core/src/configmanager.rs
```diff
@@ -11,13 +11,14 @@ use std::cmp::Ordering;
 use std::future::Future;
 use std::ops::Deref;
 use std::sync::atomic::AtomicBool;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, MutexGuard};
 use std::time::Duration;
+use tokio::sync::broadcast::error::RecvError;
 use tokio::sync::broadcast::{Receiver, Sender};
```
Peter-Searby: Considering there's two kinds of `Receiver`/`Sender` used in this module, it probably would make the code easier to follow if all those usages were namespaced, i.e. `broadcast::Receiver` and `watch::Receiver`, rather than namspacing the `watch` ones and using `Receiver`/`Sender` for the `broadcast` ones

sdk/couchbase-core/src/configmanager.rs
```diff
@@ -200,20 +193,58 @@ impl<M: KvClientManager + 'static> ConfigManagerMemdInner<M> {
 
     fn can_update_config(new_config: &ParsedConfig, old_config: &ParsedConfig) -> bool {
         if new_config.bucket != old_config.bucket {
```
Peter-Searby, line 193: I missed this on my first review, but this logic is interesting. It seems like to me that if anything in the bucket config changes (e.g. the vbucket map during a rebalance, although the debug message only mentions "bucket type"?) then it will treat this as the latest config, even if the revision is older? Also the debug message mentions "bucket "takeover", just wondering what is meant by this phrase?

sdk/couchbase-core/src/configmanager.rs
Peter-Searby: The reason for my suggestion of watch is that you can use [`Sender::send_if_modified`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_if_modified) to avoid needing to keep a dummy receiver around:
> Additionally, this method permits sending values even when there are no receivers.

sdk/couchbase-core/src/configmanager.rs
chvck: Filed RSCBC-146, this should only be checking bucket_types don't match. Bucket takeover means that select bucket has been called against a previously "bucketless" agent.

sdk/couchbase-core/src/configmanager.rs
chvck: I think `send_replace` is probably more suitable, due to us only ever calling the function if the value has changed.

sdk/couchbase-core/src/configmanager.rs
chvck: Fair point. Done.

sdk/couchbase-core/src/configmanager.rs
```diff
@@ -200,20 +192,56 @@ impl<M: KvClientManager + 'static> ConfigManagerMemdInner<M> {
...
+        tokio::spawn(async move {
+            loop {
+                match rx.recv().await {
+                    Ok(cfg) => {
+                        if let Some(new_cfg) = Self::handle_config(
+                            guard.lock().unwrap(),
+                            cfg,
+                            latest_version_tx.clone(),
+                        ) {
+                            new_config_tx.send_replace(new_cfg);
```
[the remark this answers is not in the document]
chvck: This is incorrect.

sdk/couchbase-core/src/configmanager.rs
```diff
@@ -55,19 +55,20 @@ pub(crate) struct ConfigManagerMemdInner<M: KvClientManager> {
...
-    watcher_shutdown_tx: Sender<()>,
+    on_shutdown_rx: broadcast::Receiver<()>,
+    watcher_shutdown_tx: broadcast::Sender<()>,
 }
 
 impl<M: KvClientManager + 'static> ConfigManagerMemdInner<M> {
-    pub fn watch(&self) -> Receiver<ParsedConfig> {
-        self.watcher.watch(self.watcher_shutdown_tx.subscribe())
+    pub fn watch(&self) -> watch::Receiver<ParsedConfig> {
+        self.on_new_config_tx.subscribe()
```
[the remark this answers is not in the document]
chvck: This is incorrect.

sdk/couchbase-core/src/configmanager.rs
```diff
@@ -243,33 +271,35 @@ impl<M: KvClientManager + 'static> ConfigManagerMemd<M> {
         let (watcher_shutdown_tx, watcher_shutdown_rx) = broadcast::channel(1);
         let bucket = opts.first_config.bucket.clone();
 
+        let (on_new_config_tx, on_new_config_rx) =
```
Peter-Searby: Now that `send_replace` is used, this dummy receiver should be able to be immediately dropped

Peter-Searby (approved):

PR #324, opened 2025-09-04
RSCBC-131: Make all bool option field setters take a bool param

willbroadbelt:


sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -203,13 +203,13 @@ impl KvOptions {
         Self::default()
     }
 
-    pub fn disable_mutation_tokens(mut self) -> Self {
-        self.enable_mutation_tokens = Some(false);
+    pub fn disable_mutation_tokens(mut self, disable: bool) -> Self {
+        self.enable_mutation_tokens = Some(!disable);
```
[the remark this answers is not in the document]
willbroadbelt: Agreed

sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -203,13 +203,13 @@ impl KvOptions {
...
         self
     }
 
-    pub fn disable_server_durations(mut self) -> Self {
-        self.enable_server_durations = Some(false);
+    pub fn disable_server_durations(mut self, disable: bool) -> Self {
+        self.enable_server_durations = Some(!disable);
```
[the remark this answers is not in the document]
willbroadbelt: Agreed

sdk/couchbase-core/src/options/agent.rs
```diff
@@ -193,13 +193,13 @@ impl KvConfig {
         Self::default()
     }
 
-    pub fn disable_mutation_tokens(mut self) -> Self {
-        self.enable_mutation_tokens = false;
+    pub fn enable_mutation_tokens(mut self, enable: bool) -> Self {
```
chvck, line 196: This is inverting the default value applied to the SDK, and means that by default mutation tokens will be disabled.

sdk/couchbase-core/src/options/agent.rs
```diff
@@ -193,13 +193,13 @@ impl KvConfig {
...
+        self.enable_mutation_tokens = enable;
         self
     }
 
-    pub fn disable_server_durations(mut self) -> Self {
-        self.enable_server_durations = false;
+    pub fn enable_server_durations(mut self, enable: bool) -> Self {
```
chvck, line 201: See above.

sdk/couchbase-core/src/options/agent.rs
willbroadbelt: Its just the name of the setter, KvConfig is holding 'enable_mutation_tokens' and the default is still the same. Happy to change it back (with the added bool param), most other SDKs use enableMutationTokens as the option, though guess this is core option so could go either way - I just changed for consistency.

PR #339, opened 2025-09-11
RSCBC-162: Run automated tests against 8.0

chvck:


sdk/couchbase-core/tests/search_indexes.rs
```diff
@@ -50,8 +50,7 @@ fn test_basic_search_index_management() {
 
         assert_eq!(index.name, actual_index.name);
         assert_eq!(index.index_type, actual_index.index_type);
-        assert_eq!(index.params, actual_index.params);
-        assert_eq!(index.plan_params, actual_index.plan_params);
+        // We don't test params or plan_params as the contents is server version dependent.
```
Peter-Searby, line 53: Do we not have version-dependent test handling?
Obviously we have feature-dependent handling, but not testing something just because it depends on the server version isn't a great reason

chvck: We do but params and plan_params are effectively just almost like opaque json blobs anyway, what they actually contain isn't really important to SDK - and it could change every server patch release in theory which would become a burden to manage.

PR #340, opened 2025-09-15
RSCBC-163: Add remaining transcoders

Matt-Woz:


sdk/couchbase/src/transcoding/raw_string.rs
```diff
@@ -0,0 +1,15 @@
+use crate::transcoding::{decode_common_flags, encode_common_flags, DataType};
+
+pub fn encode(value: &str) -> crate::error::Result<(&[u8], u32)> {
+    Ok((value.as_bytes(), encode_common_flags(DataType::String)))
+}
+
+pub fn decode(value: &[u8], flags: u32) -> crate::error::Result<String> {
```
chvck: Guessing we can't return a &str here? We could tie the lifetime to the lifetime of value so might work.

Matt-Woz: Good point, converted to &str

PR #341, opened 2025-09-15
RSCBC-166: Handle expiry values of > 30 days

chvck:


sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -46,6 +48,22 @@ impl CouchbaseCoreKvClient {
         }
     }
 
+    fn expiry_to_seconds(expiry: Duration) -> u32 {
```
Matt-Woz: Should we log a warning if the result is > `u32::MAX` ?

chvck: I'm inclined to say no. If anything we should return an error but unsure if it's necessary. Maybe.

Matt-Woz: I was also thinking an error could be suitable but a warning seems like an OK middle-ground, I don't think we should just be accepting it silently, though?

Matt-Woz (approved):

PR #360, opened 2025-09-23
RSCBC-182: Update TlsOptions to take a vec of certificates

Matt-Woz:


sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -293,14 +293,14 @@ impl TlsOptions {
     }
 
     #[cfg(all(feature = "rustls-tls", not(feature = "native-tls")))]
-    pub fn ca_certificate(mut self, cert: CertificateDer<'static>) -> Self {
-        self.ca_certificate = Some(cert);
+    pub fn add_ca_certificate(mut self, cert: CertificateDer<'static>) -> Self {
```
chvck, line 296: We should probably add a function which accepts a vec of these all at once too. That would probably also necessitate making certs an option so we can detect the user giving us an empty vec.

PR #361, opened 2025-09-23
RSCBC-183: AtPlus ScanConsistency should be updated to contain a mutation state

Matt-Woz:


sdk/couchbase/src/options/query_options.rs
```diff
@@ -231,14 +225,13 @@ impl TryFrom<QueryOptions> for query::QueryOptions {
     type Error = error::Error;
 
     fn try_from(opts: QueryOptions) -> Result<query::QueryOptions, Self::Error> {
-        let (mutation_state, scan_consistency) = if let Some(mutation_state) = opts.consistent_with
-        {
-            (
-                Some(mutation_state.into()),
+        let (mutation_state, scan_consistency) = match &opts.scan_consistency {
```
chvck: Does this need to be a borrow?

Matt-Woz: Nope, fixed

sdk/couchbase/src/options/query_options.rs
```diff
@@ -231,14 +225,13 @@ impl TryFrom<QueryOptions> for query::QueryOptions {
...
-                Some(mutation_state.into()),
+        let (mutation_state, scan_consistency) = match opts.scan_consistency {
+            Some(ScanConsistency::AtPlus(state)) => (
+                Some(state.clone().into()),
                 Some(queryx::query_options::ScanConsistency::AtPlus),
-            )
-        } else {
-            (None, opts.scan_consistency.map(|sc| sc.into()))
+            ),
+            Some(scan_consistency) => (None, Some(scan_consistency.clone().into())),
```
chvck: Does this need to be clone now? Same above.

sdk/couchbase/src/options/query_options.rs
```diff
@@ -231,14 +225,13 @@ impl TryFrom<QueryOptions> for query::QueryOptions {
...
-                Some(mutation_state.into()),
+        let (mutation_state, scan_consistency) = match opts.scan_consistency {
+            Some(ScanConsistency::AtPlus(state)) => (
+                Some(state.into()),
                 Some(queryx::query_options::ScanConsistency::AtPlus),
-            )
-        } else {
-            (None, opts.scan_consistency.map(|sc| sc.into()))
+            ),
+            Some(scan_consistency) => (None, Some(scan_consistency.into())),
```
chvck: I wonder if we should remove the From impl for ScanConsistency and just match here? It feels like a bug in waiting right now.

Matt-Woz: Sure

sdk/couchbase/src/options/query_options.rs
```diff
@@ -8,20 +8,20 @@ use serde_json::Value;
...
+#[derive(Debug, Clone, Eq, PartialEq)]
 #[non_exhaustive]
 pub enum ScanConsistency {
     NotBounded,
     RequestPlus,
-    AtPlus,
+    AtPlus(MutationState),
 }
 
 impl From<ScanConsistency> for queryx::query_options::ScanConsistency {
```
chvck: We should remove this so that no-one can accidentally use it.

chvck (approved):

PR #363, opened 2025-09-24
RSCBC-188: Can't add search query to request if vector set

Matt-Woz:


Matt-Woz:
> I do wonder if we should remove these checks so we don't return Result but +2

Yeah idk, I think it would be better UX but the RFC is very prescriptive with this:

> - searchQuery(SearchQuery searchQuery). If the user has already specified a SearchQuery, either at construction time or via another call to SearchRequest.searchQuery(), the SDK needs to raise an InvalidArgumentException.
> - vectorSearch(VectorSearch vectorSearch). If the user has already specified a VectorSearch, either at construction time or via another call to SearchRequest.vectorSearch(), the SDK needs to raise an InvalidArgumentException.

Issue #367, opened 2025-09-25
Check connections when client pool cannot find any client to return


gu0keno0:
The current OSS couchbase client only try to recreate connection once when the connection to the host is closed. If that retry fails, all the get() calls on connection pool returns cached error in the state object.

Repro:

iptables -I OUTPUT 1 -p tcp --dport 11207 -j REJECT --reject-with tcp-reset

sleep 60

iptables -D OUTPUT -p tcp --dport 11207 -j REJECT --reject-with tcp-reset

# The Rust CB client can no longer recover because of the issue.

gu0keno0:
https://github.com/couchbaselabs/couchbase-rs/pull/368/files is our attempt to address this internally. Tested by both the above case and the general prod usage.

chvck:
Looking at the code it looks like the SDK should already be retrying the connection after it errors, indefinitely. We will be returning the cached error until a connection attempt succeeds though. I'll try to get your repro working in my environment.

gu0keno0:
It seems that after blocking all 11207 ports with REJECT, the reconnection attempt stops. I think it's because REJECT with RST can trigger graceful connection close on Linux, and because of that, the regular connection retries that works for errors stops.

chvck:
I've tried running the same on my system and every time I see the SDK recover after I delete the iptables rule. I'll keep looking into trying different scenarios.

PR #368, opened 2025-09-25
Check connections when get_client_slow fails

gu0keno0:
Fix for https://github.com/couchbaselabs/couchbase-rs/issues/367

### DIFF
--- a/sdk/couchbase-core/src/kvclientpool.rs
+++ b/sdk/couchbase-core/src/kvclientpool.rs
@@ -163,7 +163,19 @@ where
             }
         }
 
-        self.get_client_slow().await
+        match self.get_client_slow().await {
+            Ok(c) => Ok(c),
+            Err(e) => {
+                // If we failed to get a client, we should check connections and rebuild the fast map.
+                // For example, when all clients are terminated by server side sending TCP RST,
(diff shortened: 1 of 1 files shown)


sdk/couchbase-core/src/kvclientpool.rs
```diff
@@ -163,7 +163,19 @@ where
             }
         }
 
-        self.get_client_slow().await
+        match self.get_client_slow().await {
+            Ok(c) => Ok(c),
+            Err(e) => {
```
chvck, line 168: This will end up attempting to rebuild connections when get_client_slow returns Shutdown. I think that the correct approach here might be to not check for a cached error at all in get_client_slow, and to call 

```
self.check_connections().await;
self.rebuild_fast_map().await;
```

either at that point or do something similar when client creation fails within `start_new_client_thread` so that we proactively attempt to have a new client ready.  What do you think?

As an aside between our developer preview and GA release I'm hoping to rewrite the kvclient, kvclientpool, and kvclientmanager stack as it's too complicated and difficult to understand (especially failure cases)

gu0keno0: Fair point. I'll look into the two approaches and update the PR.

PR #369, opened 2025-09-26
RSCBC-191: Make error::new public

Matt-Woz:


sdk/couchbase/src/error.rs
```diff
@@ -18,11 +18,41 @@ pub struct Error {
...
     pub(crate) fn new(kind: ErrorKind) -> Self {
-        Self {
-            kind: Box::new(kind),
-            context: Box::new(None),
-        }
+        Self::new_internal(kind)
+    }
+
+    #[cfg(feature = "unstable")]
+    pub fn invalid_argument(arg: impl Into<String>, msg: impl Into<String>) -> Self {
```
chvck: I think we should probably make these not public and just make invalids args/feature not available error types public. Although I'm not certain on that...

Matt-Woz: Yeah I'm not sure either, though it's unlikely to change I feel cautious of exposing the struct directly without #[non_exhaustive] just in case, which requires a constructor anyway, and this way we still have its construction marked as unstable

sdk/couchbase/Cargo.toml
```diff
@@ -39,8 +39,8 @@ default = ["default-tls"]
 default-tls = ["rustls-tls"]
 rustls-tls = ["dep:tokio-rustls", "couchbase-core/rustls-tls", "dep:rustls-pemfile"]
 native-tls = ["dep:tokio-native-tls", "couchbase-core/native-tls"]
-uncommitted = []
-volatile = []
+unstable = []
```
chvck: Is it worth backing this change out given the other PR?

Matt-Woz: I will rebase once that's merged & change this to the new style

Matt-Woz: Done

PR #372, opened 2025-10-01
RSCBC-198: Rework unstable feature flagging

Matt-Woz:


sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -99,7 +99,8 @@ impl ClusterOptions {
         self
     }
 
-    #[cfg(feature = "volatile")]
+    /// Volatile: This API is subject to change at any time
```
chvck: I think this comment should actually go on the feature in the Cargo.toml, so that it's really clear to users. Probably with a brief description what the options do.

Matt-Woz: Done

PR #374, opened 2025-10-02
RSCBC-110: Update options to make fields public

willbroadbelt:


sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -410,7 +413,7 @@ impl TlsOptions {
     }
 
     #[cfg(feature = "native-tls")]
-    pub(crate) fn try_into_tls_config(
+    pub fn try_into_tls_config(
```
Matt-Woz: Does this have to be public?

sdk/couchbase/src/search/vector.rs
```diff
@@ -4,8 +4,8 @@ use couchbase_core::searchx::query_options::{KnnOperator, KnnQuery};
 
 #[derive(Debug, Clone)]
 pub struct VectorSearch {
```
Matt-Woz, line 7: We should probably make this #[non_exhaustive]

sdk/couchbase/src/options/cluster_options.rs
willbroadbelt: No true, done.

sdk/couchbase/src/search/vector.rs
willbroadbelt: Done

Matt-Woz (approved):

PR #379, opened 2025-10-08
RSCBC-196: Prepare v1.0.0-beta.1 release

chvck:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,22 @@
+name: Publish crate
```
chvck, line 1: This action is going to fall short as soon as we need to publish any individual crate "out of band" but we can cross that bridge when we get to it.

PR #383, opened 2025-10-13
Community version testing

chvck:


.github/workflows/tests.yml
```diff
@@ -14,19 +14,27 @@ jobs:
...
+          - server: 7.6.7
+            version: 7.6.0
+          - server: 7.2.7
+            version: 7.2.0
+          - server: 7.1.6
+            version: 7.1.0
+          - server: 7.0.4
+            version: 7.0.0
+          - server: community-7.6.2
+            version: 7.6.0-community
```
willbroadbelt: Do we need to have these separate server/versions? There are tests specifying patches i.e. SERVER_VERSION_762 that wont get run like this when they should be. Feels like modifying the rust test config to recognise 'community-7.6.2' and '7.6.0-community' formats would simplify things.

.github/workflows/tests.yml
```diff
@@ -50,12 +58,22 @@ jobs:
...
+        run: |
+          CBDC_CONNSTR=$(cbdinocluster -v connstr $CBDC_ID)
+          echo "CBDC_CONNSTR=$CBDC_CONNSTR" >> "$GITHUB_ENV"
+        if: matrix.server == 'community-7.6.2'
+
+      - name: Set connection string
+        run: |
+          CBDC_CONNSTR=$(cbdinocluster -v connstr --tls $CBDC_ID)
+          echo "CBDC_CONNSTR=$CBDC_CONNSTR" >> "$GITHUB_ENV"
+        if: matrix.server != 'community-7.6.2'
```
willbroadbelt: Could combine these into one step inside the run shell script. Could be handy in the future if wanted an option to run with/without TLS independent of it being community/enterprise. 

Can also match more generically on community edition rather than the specific community version, something like:
if [[ ${{ matrix.server }} == comunity* ]]; then

chvck: Done

.github/workflows/tests.yml
chvck: This is not nearly as simple as it sounds to do so I'm very hesitant to do it.

willbroadbelt (approved):

PR #384, opened 2025-10-15
RSCBC-213: Don't allocate when encoding framing extras

chvck:


sdk/couchbase-core/tests/agent_ops.rs
```diff
@@ -809,9 +809,9 @@ fn upsert_allocations() {
         let total_allocs = stats2.total_blocks - stats1.total_blocks;
 
         let expected_allocs = if agent.test_setup_config.use_ssl {
```
chvck, line 811: These were backward.

PR #388, opened 2025-10-22
RSCBC-136: Rework kv client stack

chvck:


sdk/couchbase-core/src/agent.rs
```diff
@@ -144,19 +149,24 @@ pub struct Agent {
...
-    pub query_config: QueryComponentConfig,
-    pub search_config: SearchComponentConfig,
-    pub mgmt_config: MgmtComponentConfig,
-    pub diagnostics_config: DiagnosticsComponentConfig,
-    pub http_client_config: ClientConfig,
-}
-
 impl AgentInner {
-    pub async fn unsolicited_packet_handler(&self, packet: ResponsePacket) {
+    fn gen_agent_component_configs_locked(state: &AgentState) -> AgentComponentConfigs {
```
Matt-Woz, line 153: Would it make sense for this to be a method of AgentState?

chvck: I don't think it can be, as it updates all of the components that are a part of AgentInner too.

PR #389, opened 2025-10-22
RSCBC-72: Add Projections using Get

willbroadbelt:
Motivation:
Add the Projections using Get feature, as defined in RFC #53.

Changes:
- Add projections Option to GetOptions
- Do full doc Get or subdoc Lookup In depending on if projections and/or expiry is defined, and within lookup in limits (16).
- Add test cases

### DIFF
--- a/sdk/couchbase/src/clients/couchbase_core_kv_client.rs
+++ b/sdk/couchbase/src/clients/couchbase_core_kv_client.rs
@@ -30,6 +30,7 @@ use crate::results::kv_results::{
     ExistsResult, GetResult, LookupInResult, LookupInResultEntry, MutateInResult,
     MutateInResultEntry, MutationResult, TouchResult,
 };
+use crate::results::projection::{build_from_full_doc, build_from_subdoc_entries};
 use crate::subdoc::lookup_in_specs::{GetSpecOptions, LookupInSpec};
 use crate::subdoc::mutate_in_specs::MutateInSpec;
 use chrono::{DateTime, Utc};
@@ -221,35 +222,15 @@ impl CouchbaseCoreKvClient {
     }
 
     pub async fn get(&self, id: &str, options: GetOptions) -> error::Result<GetResult> {
-        let agent = self.agent_provider.get_agent().await;
-
-        if let Some(true) = options.expiry {
-            let specs = vec![
-                LookupInSpec::get("$document.exptime", GetSpecOptions::new().xattr(true)),
-                LookupInSpec::get("$document.flags", GetSpecOptions::new().xattr(true)),
-                LookupInSpec::get("", None),
-            ];
-
-            let res = self.lookup_in(id, &specs, LookupInOptions::new()).await?;
-            let expiry: u64 = res.content_as(0)?;
-            let expires_at = match DateTime::<Utc>::from_timestamp(expiry as i64, 0) {
-                Some(e) => e,
-                None => {
-                    return Err(error::Error::other_failure(
-                        "invalid expiry time returned from server".to_string(),
-                    ));
-                }
-            };
-            let flags: u32 = res.content_as(1)?;
-            let content: Vec<u8> = res.content_as_raw(2)?.to_vec();
-
-            return Ok(GetResult {
-                content,
-                flags,
-                cas: res.cas,
-                expiry_time: Some(expires_at),
-            });
+        if options.expiry.is_some() || options.projections.is_some() {
+            self.get_with_projections(id, options).await
+        } else {
+            self.get_direct(id, options).await
         }
+    }
+
+    async fn get_direct(&self, id: &str, options: GetOptions) -> error::Result<GetResult> {
+        let agent = self.agent_provider.get_agent().await;
 
         let res = CouchbaseAgentProvider::upgrade_agent(agent)?
             .get(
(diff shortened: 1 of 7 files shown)


sdk/couchbase/tests/common/doc_generation.rs
```diff
@@ -108,6 +109,62 @@ pub async fn import_sample_beer_dataset(
...
+pub async fn import_projection_doc(collection: &Collection) -> String {
+    let content = read_to_string("tests/testdata/projection_doc.json").unwrap();
+    let doc: serde_json::Value = serde_json::from_str(content.as_str()).unwrap();
+
+    let key = new_key();
+    collection.upsert(key.clone(), &doc, None).await.unwrap();
+    key
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, PartialOrd, Default)]
```
chvck, line 121: Do all of these throughout this JSON type need to be optional fields? Certainly name is always set.

sdk/couchbase/tests/collection_crud.rs
```diff
@@ -565,3 +568,408 @@ fn get_with_expiry() {
...
+            .unwrap();
+
+        #[derive(Debug)]
+        struct TCase {
+            name: &'static str,
+            project: Vec<String>,
+            expected: Person,
+        }
+
+        let cases: Vec<TCase> = vec![
```
chvck, line 594: Unsure about this, if any of these cases fail then it'll halt the execution of the rest. Which could be annoying?

sdk/couchbase/src/results/projection.rs
```diff
@@ -0,0 +1,223 @@
...
+        ));
+    }
+
+    let content: Value = result.content_as(0)?;
+    let mut projection_content: Value = Value::Object(Default::default());
+    for projection in projections.unwrap() {
+        let parts = SubdocPath::path_parts(projection.as_str());
+        if let Some(value) = content.get(projection).cloned() {
+            projection_content = SubdocPath::set_value(&parts, projection_content, value);
+        } else {
```
chvck, line 81: If a projection is path shouldn't we return an error?

sdk/couchbase/src/results/projection.rs
```diff
@@ -0,0 +1,223 @@
...
+            parts.push(SubdocPath::new(current, false));
+        }
+
+        parts
+    }
+
+    fn set_value_from_bytes(paths: &[SubdocPath], content: Value, value_bytes: &[u8]) -> Value {
+        // Try to parse bytes into a JSON Value
+        let value: Value = from_slice(value_bytes).unwrap_or_else(|_| {
+            error!("Warning: value was not valid JSON, starting with empty object.");
```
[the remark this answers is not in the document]
chvck: This should just return an error I think.

sdk/couchbase/src/results/kv_results.rs
```diff
@@ -141,17 +141,17 @@ impl MutationResult {
     }
 }
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub(crate) struct LookupInResultEntry {
-    pub(crate) value: Option<Bytes>,
-    pub(crate) error: Option<error::Error>,
-    pub(crate) op: LookupInOpType,
+pub struct LookupInResultEntry {
```
chvck: Why have we made these public? I think we generally want to keep result fields private.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -265,6 +246,90 @@ impl CouchbaseCoreKvClient {
...
+                ));
+            }
+        }
+
+        if projections.is_empty() {
+            specs.push(LookupInSpec::get("", None));
+        } else {
+            for projection in projections.clone() {
+                specs.push(LookupInSpec::get(projection, None));
+                num_projections += 1;
```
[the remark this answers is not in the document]
willbroadbelt: Changed name to be more clear.

sdk/couchbase/src/results/projection.rs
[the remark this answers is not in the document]
willbroadbelt: Done.

sdk/couchbase/tests/common/doc_generation.rs
willbroadbelt: Thought I needed it to parse in the projection gets into this struct - which miss fields, but can mark it #[serde(default)] which is better. done.

sdk/couchbase/tests/collection_crud.rs
willbroadbelt: True, I've made it collect up the results and report the failures at the end instead of failing immediately for a TC.

sdk/couchbase/src/results/kv_results.rs
willbroadbelt: Whoops left in before I refactored some things - reverted.

sdk/couchbase/src/results/projection.rs
```diff
@@ -0,0 +1,223 @@
...
+    let mut projection_content: Value = Value::Object(Default::default());
+    for projection in projections.unwrap() {
+        let parts = SubdocPath::path_parts(projection.as_str());
+        if let Some(value) = content.get(projection).cloned() {
+            projection_content = SubdocPath::set_value(&parts, projection_content, value);
+        } else {
+            warn!(
+                "projection path '{}' not found in document; skipping",
+                projection
+            );
```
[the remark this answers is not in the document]
willbroadbelt: Done.

sdk/couchbase/src/results/projection.rs
willbroadbelt: If a given projection path returns an error its silently omitted from the results, is what the RFC states to do - if thats the question?

chvck: Fair enough

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -265,6 +246,90 @@ impl CouchbaseCoreKvClient {
...
 
+    async fn get_with_projections(
+        &self,
+        id: &str,
+        options: GetOptions,
+    ) -> error::Result<GetResult> {
+        let mut num_lookups = options.projections.clone().unwrap_or_default().len();
+        let mut with_flags = false;
+
+        if options.expiry.is_some() {
```
chvck: I think this should possible be `   if let Some(true) = options.expiry {`?

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -265,6 +246,90 @@ impl CouchbaseCoreKvClient {
         Ok(res.into())
     }
 
+    async fn get_with_projections(
+        &self,
+        id: &str,
+        options: GetOptions,
+    ) -> error::Result<GetResult> {
+        let mut num_lookups = options.projections.clone().unwrap_or_default().len();
```
chvck: There must be a way that we can refactor this to not clone here and on L264. The unwrap_or_default is a bit suspicious on them too.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -265,6 +246,90 @@ impl CouchbaseCoreKvClient {
...
+        let mut with_flags = false;
+
+        if options.expiry.is_some() {
+            if num_lookups == 0 {
+                with_flags = true;
```
chvck, line 260: It might be useful to have some comments around why with_flags is a thing.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -265,6 +246,90 @@ impl CouchbaseCoreKvClient {
...
+            }
+            num_lookups += 1;
+        }
+
+        let mut projections = options.projections.clone().unwrap_or_default();
```
chvck: This clone is very suspicious. We appear to clone projections multiple times in this function.

sdk/couchbase/src/results/mod.rs
```diff
@@ -20,6 +20,7 @@ pub mod collections_mgmt_results;
 pub mod diagnostics;
 pub mod kv_binary_results;
 pub mod kv_results;
+pub mod projection;
```
chvck: pub(crate). In another change we need to go through and fix that for another other modules that don't actually export anything public.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
willbroadbelt: Done.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
willbroadbelt: True, added.

sdk/couchbase/src/results/mod.rs
willbroadbelt: Done.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
willbroadbelt: Good point - updated.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
willbroadbelt: As above, changed how I've done this a bit and removed the clones.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -265,6 +246,92 @@ impl CouchbaseCoreKvClient {
...
+            }
+        }
+
+        if full_lookup {
+            specs.push(LookupInSpec::get("", None));
+        } else {
+            for projection in options.projections.as_deref().unwrap() {
+                specs.push(LookupInSpec::get(projection, None));
+                num_lookups += 1;
+            }
```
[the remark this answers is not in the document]
willbroadbelt: No, we're counting the expiry lookupIn on purpose.

sdk/couchbase/src/clients/couchbase_core_kv_client.rs
```diff
@@ -265,6 +246,92 @@ impl CouchbaseCoreKvClient {
         Ok(res.into())
     }
 
+    async fn get_with_projections(
+        &self,
+        id: &str,
+        options: GetOptions,
+    ) -> error::Result<GetResult> {
+        let mut num_lookups = options.projections.as_deref().map_or(0, |p| p.len());
```
chvck, line 254: I think this function can be rewritten to be more rust-y but let's get it in for now with the tests.

PR #409, opened 2026-01-13
RSCBC-88: Add retry strategy back into public API

chvck:


sdk/couchbase-core/src/retry.rs
```diff
@@ -164,38 +125,31 @@ impl RetryAction {
...
 
 #[derive(Clone, Debug)]
-pub struct RetryInfo {
-    pub(crate) operation: &'static str,
-    pub(crate) is_idempotent: bool,
-    pub(crate) retry_strategy: Arc<dyn RetryStrategy>,
-    pub(crate) retry_attempts: u32,
-    pub(crate) retry_reasons: HashSet<RetryReason>,
-    pub(crate) unique_id: Option<String>,
+pub struct RetryRequest {
```
Matt-Woz, line 133: Out of curiosity, what is the motivation for extracting out the `retry_strategy` from here?

Matt-Woz (approved):

sdk/couchbase-core/src/retry.rs
chvck:
```
pub trait RetryStrategy: Debug + Send + Sync {
    fn retry_after(&self, request: &RetryRequest, reason: &RetryReason) -> Option<RetryAction>;
}
```
It's very weird for RetryStrategy to accept a parameter which is a type which contains RetryStrategy.

Matt-Woz (approved):

PR #418, opened 2026-01-28
RSCBC-236: Log cluster and agent information at startup

chvck:


sdk/couchbase/src/options/cluster_options.rs
```diff
@@ -194,6 +208,17 @@ impl From<PollerOptions> for couchbase_core::options::agent::ConfigPollerConfig
     }
 }
 
+// Display for PollerOptions
```
Matt-Woz: tiny NIT: these comments

Matt-Woz (dismissed):

sdk/couchbase/src/options/cluster_options.rs
chvck: Because LLM

Issue #480, opened 2026-05-13
Allow for more control over dependencies


rukai:
Hi I'm trying to improve incremental compilation times of my application.
I can see that couchbase is pulling in a lot of dependencies that are included in my application but not actually used by it.

Currently `couchbase` pulls in the following transitive dependencies that I would like to see made optional:
* hickory-resolver - this is a really large dependency for an optional feature, would be good to put behind a feature flag
* tracing-subscriber - tracing-subscriber is generally a dependency for applications only, so whatever functionality relies on it could probably go behind a feature flag
* hdrhistogram - another fairly big dependency, the logging_meter stuff looks like it could easily go behind a feature flag.

You can still keep the library easy to use by including all of these features in a default feature flag.
And you can use a tool like cargo hack in your CI to ensure that all combinations of features can build correctly.

I'm also not sure of your MSRV situation but you could get a small but easy win from replacing lazy_static with the std library alternatives.

Thanks!

chvck:
Hi @rukai thanks for raising this, this seems like a reasonable request to me. I've raised https://jira.issues.couchbase.com/browse/RSCBC-275 for internal tracking purposes.

rukai:
This is great, thankyou!
Measuring exact warm build time improvements is a pain, but using the size of the debug build artifact as a proxy I observed that this change reduced the size of our projects debug build binary from 133MB to 125MB, a solid win for our moderately sized project.

I'll leave the ticket open in case you want to track it to release, but feel free to close now.

chvck:
Thanks for confirming @rukai, will close now.

PR #482, opened 2026-05-21
RSCBC-275: Allow for more control over dependencies

Matt-Woz:
Addresses issue #480

### DIFF
--- a/sdk/couchbase-connstr/src/lib.rs
+++ b/sdk/couchbase-connstr/src/lib.rs
@@ -19,19 +19,17 @@
 pub mod error;
 
 use error::ErrorKind;
-use hickory_resolver::config::*;
-use hickory_resolver::name_server::TokioConnectionProvider;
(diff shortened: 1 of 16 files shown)


sdk/couchbase-connstr/src/lib.rs
```diff
@@ -19,17 +19,25 @@
 pub mod error;
 
 use error::ErrorKind;
+#[cfg(feature = "dns-srv")]
```
chvck: Is there a way for us to group these use statements that are conditional? Something like https://stackoverflow.com/questions/58167542/how-to-group-imports-in-rust-with-conditional-compilation

chvck (commented):
I suspect this should probably be part of a dotminor release?

sdk/couchbase-connstr/src/lib.rs
Matt-Woz: Done

chvck (approved):

PR #489, opened 2026-07-14
RSCBC-281: Erroneous log statement on shutdown

Matt-Woz:


chvck (commented):
I might be missing something but where is closed set to true?

Matt-Woz:
> I might be missing something but where is closed set to true?

It is set in the Client's `close()` function, which wasn't modified in this change

chvck (approved):

PR #493, opened 2026-07-16
RSCBC-285: create_index cannot index nested fields properly

Matt-Woz:
Motivation
-----------
Currently the create_index escapes the fields that are passed in, this means that a field like `d.firstName` was parsed as '\`d.firstName\`' and not '\`d\`.\`firstName\`'. This meant that it was not possible to index nested fields properly. Likewise in the case when backticks were needed in the field name itself, escaping the backticks would cause the server to return a syntax error.

Changes
----------
Stop escaping index field names and pass them as is to allow indexing nested fields

Notes
-------
This change may break anyone who was relying on the SDK escaping the index names, e.g. trying to index non-nested fields with dots in their name, or when reserved words are used, in which case backticks will have to be added to the user's field input.

### DIFF
--- a/sdk/couchbase-core/src/queryx/query.rs
+++ b/sdk/couchbase-core/src/queryx/query.rs
@@ -348,11 +348,7 @@ impl<C: Client> Query<C> {
             build_keyspace(opts.bucket_name, &opts.scope_name, &opts.collection_name)
         ));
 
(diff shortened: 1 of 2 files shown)


chvck (commented):
We should probably have a test for this

chvck (approved):

PR #494, opened 2026-08-04
RSCBC-288: SDK Support for Hybrid Search Score Fusion

Matt-Woz:


Matt-Woz:
Leaving as draft until the FIT gRPC is merged.

sdk/couchbase/src/options/search_options.rs
```diff
@@ -149,11 +150,16 @@ pub struct SearchOptions {
     pub sort: Option<Vec<Sort>>,
     /// Facets to include in the result for aggregation.
     pub facets: Option<HashMap<String, Facet>>,
+    /// Controls how results are scored. See [`Scoring`] for the available modes.
+    ///
+    /// **Volatile: This API is subject to change at any time.**
```
chvck: Presumably this won't be volatile by the time it's in a release? It's a bit odd to deprecate disable_scoring and mark this as volatile as it leaves the user with no stable API to disable scoring.

sdk/couchbase/src/clients/search_client.rs
```diff
@@ -113,14 +115,43 @@ impl CouchbaseSearchClient {
...
-            }
+        if opts.disable_scoring.is_some() && opts.scoring.is_some() {
+            return Err(error::Error::invalid_argument(
+                "scoring",
+                "cannot be used with disable_scoring",
+            ));
+        }
+
+        let (score, params) = if let Some(scoring) = opts.scoring {
+            let score = match &scoring {
```
chvck: Why don't we just collapse these 2 matches into one?

sdk/couchbase/src/options/search_options.rs
Matt-Woz: True, removed the volatile tag

sdk/couchbase/src/clients/search_client.rs
Matt-Woz: Good point, done

chvck (approved):

PR #505, opened 2026-09-09
RSCBC-301: SDK doesn't send name during memcached HELO

Matt-Woz:


sdk/couchbase-core/src/memdx/ops_core.rs
```diff
@@ -102,6 +102,12 @@ impl OpBootstrapEncoder for OpsCore {
             features.extend_from_slice(&bytes);
         }
 
+        let key = if request.client_name.is_empty() {
+            None
+        } else {
+            Some(request.client_name.as_slice())
```
chvck, line 108: Do we need to make sure this is under 200 characters somewhere?

sdk/couchbase-core/src/agent.rs
```diff
@@ -351,7 +352,7 @@ impl ConfigUpdater for AgentInner {
 impl Agent {
     pub async fn new(opts: AgentOptions) -> Result<Self> {
         let build_version = env!("CARGO_PKG_VERSION");
-        let user_agent = format!("cb-rust/{build_version}");
+        let user_agent = format!("cb-rust/{build_version} ({} {})", OS, ARCH);
```
chvck: Can't this just be format!("cb-rust/{build_version} ({OS} {ARCH})?

sdk/couchbase-core/src/memdx/ops_core.rs
Matt-Woz: Done

sdk/couchbase-core/src/agent.rs
Matt-Woz: Done

DemetrisChr (approved):

